PR #15, opened 2020-08-25
Fix limit local dev deployment

### LINKED ISSUE #11: Limit local dev deployments to localhost
App deploys statically to `0.0.0.0`. This should be limited for local development setups so the developer's permission set is not given a public endpoint.

wg102:
Resolves #11

### DIFF
--- a/main.go
+++ b/main.go
@@ -35,6 +35,7 @@ var kubeconfig string
 var spawnerConfigPath string
 var userIDHeader string
 var staticDirectory string
+var listenAddr string
 
 type listers struct {
 	namespaces             v1listers.NamespaceLister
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -76,6 +77,7 @@ func main() {
 	flag.StringVar(&userIDHeader, "userid-header", "kubeflow-userid", "header in the request which identifies the incoming user")
 	flag.StringVar(&spawnerConfigPath, "spawner-config", "/etc/config/spawner_ui_config.yaml", "path to the spawner configuration file")
 	flag.StringVar(&staticDirectory, "static-dir", "static/", "path to the static assets")
+	flag.StringVar(&listenAddr, "listen-addr", "127.0.0.1:5000", "server listen address")
```
brendangadd, line 80: Can we do the whole configuration fallback right here (with an extra functions)? Like...

```go
// With this function defined...
func lookupEnv(name string, defaultValue string) string {
	if value, isSet := os.LookupEnv(name); isSet {
		return value
	}
	return defaultValue
}
// You can do...
flag.StringVar(&listenAddress, "listen-address",
	lookupEnv("LISTEN_ADDRESS", "127.0.0.1:5000"),
	"host and port to bind")
```

Issue #18, opened 2020-05-11
Procedure to destroy old notebook volumes


ca-scribner:
Every notebook server by default creates a volume.  Users can optionally create additional persistent volumes at notebook creation.  These are not destroyed at notebook termination.

Possible solutions:
- do we care?  If we don't mind having these persist indefinitely, close this :)
- do we have an existing way for users to cull these.  
  - If yes: add it to docs for storage
  - If no: should we?
- do we have automated garbage detection for the system to notice when users aren't culling these themselves?

brendangadd:
As a user, I want to see all my notebook volumes and delete the ones I don't need.

Let's add UI to the Jupyter web app component of Kubeflow to enable listing and deletion of persistent volumes.
* List all of (and only) the volumes owned by the user
* Only allow deletion of volumes that are not in use – probably equivalent to having no PVC

The logic for displaying volumes and allowing deletion is probably similar to that which displays the list of existing volumes available to be attached to a new notebook server.

/cc @wg102

PR #20, opened 2020-09-11
Support notebook volume (PVC) management

wg102:
Fixes https://github.com/StatCan/jupyter-apis/issues/18

Note:
Since the data is now passed through main-table:
- makes similar changes for the resource-table
- moves the Confirmation Dialog into the main-table folder.


frontend/src/app/main-table/volumes-table/volume-table.component.html
```diff
@@ -0,0 +1,41 @@
...
+
+    <ng-container matColumnDef="namespace">
+      <th mat-header-cell *matHeaderCellDef>Namespace</th>
+      <td mat-cell *matCellDef="let elem">{{ elem.pvc.namespace }}</td>
+    </ng-container>
+
+    <ng-container matColumnDef="isMounted">
+      <th mat-header-cell *matHeaderCellDef>Is Mounted?</th>
+      <td mat-cell *matCellDef="let elem">
+        {{ true === elem.ismounted ? 'Yes' : 'No' }}
```
brendangadd, line 20: `x === true ≡ x`

frontend/src/app/main-table/volumes-table/volume-table.component.scss
```diff
@@ -0,0 +1,86 @@
+.card {
```
brendangadd, line 1: This is copied from another component. Move it up and reference it from both places.

frontend/src/app/main-table/volumes-table/volume-table.component.html
brendangadd, line 20: `x === true ≡ x`

frontend/src/app/main-table/volumes-table/volume-table.component.ts
```diff
@@ -0,0 +1,152 @@
...
+        this.poller.reset();
+      });
+
+    this.subscriptions.add(resourcesSub);
+    this.subscriptions.add(namespaceSub);
+  }
+
+  updatePvcStatus(forceRefresh:boolean){
+    //Get the list of used pvcs
+    this.k8s.getResource(this.currNamespace).subscribe(resources => {
```
brendangadd, line 90: Subs not being cleaned up on destruction.

frontend/src/app/main-table/volumes-table/volume-table.component.ts
```diff
@@ -0,0 +1,152 @@
...
+
+      //TODO: make sure the getResource was ran first
+      this.k8s.getVolumes(this.currNamespace).subscribe(pvcs => {
+        if(isEqual(this.pvcs, pvcs)){
+          return;
+        }
+        this.pvcs = pvcs;
+        this.updatePvcStatus(true); //force the update of pvc status     
+      }); 
+      this.updatePvcStatus(false);
```
brendangadd, line 73: The two requests are independent, so they can be done in parallel. If you decouple the concepts at play here, I think the implementation will become clearer:

* Requests, streams, etc. go fetch, or receive, updates in state – akin to what a controller handles in MVC
* State is passed to components for rendering as a (changing) data structure – akin to view in MVC
* View renders meaningful UI for the various possible states of its input data

So you can separate the concerns of setting the data structure on the component context from the rendering the current state of the context.

frontend/src/app/main-table/volumes-table/volume-table.component.ts
```diff
@@ -0,0 +1,152 @@
...
+        yes: "delete",
+        no: "cancel"
+      }
+    });
+
+    dialogRef
+    .afterClosed()
+    .pipe(first())
+    .subscribe(result => {
+      if (!result || result !== "delete") {
```
brendangadd, line 133: `!result || result !== "delete" ≡ result !== "delete"`

frontend/src/app/services/kubernetes.service.ts
```diff
@@ -124,6 +124,21 @@ export class KubernetesService {
     );
   }
 
+  // Delete pvc
+  deletePersistentStorageClaim(ns: string, nm: string): Observable<string> {
```
brendangadd, line 128: What is `nm`? `name`? If so, definitely worth the extra two letters.

brendangadd (changes requested):

frontend/src/app/main-table/volumes-table/volume-table.component.ts
```diff
@@ -1,21 +1,27 @@
 import {Component, Input, Output, EventEmitter} from "@angular/core";
 import {MatTableDataSource} from "@angular/material/table";
-import {Pvc} from "../../utils/types";
+import {Volume} from "../../utils/types";
+
+export type PvcWithStatus = {
+  pvc: Volume;
+  mountedBy: string | null;
+  status?: boolean;
```
brendangadd, line 8: Putting status on this type would make sense if we were following the approach of the Notebooks table and the deletion status were coming from the API server. But here we are tracking deletion status purely in the front-end, and simply for UI display purposes.

So we want to keep ownership of that information inside the display component so that other components don't have to concern themselves with managing deletion state.

All that to say, let's keep deletion status tracking inside this component.

And, once you do, make sure you clean up status tracking for volumes that no longer exist. Otherwise it will be possible to recreate a deleted volume and have it incorrectly showing as deleting in the table once it starts getting listed by the API server.

brendangadd (changes requested):

brendangadd (changes requested):
I don't see any cleanup of PVC deletion tracking on completion.

brendangadd (approved):

Issue #25, opened 2020-09-29
Display cost estimates for namespace


brendangadd:
As a user, I want to see an estimated daily cost of the resources I'm using. I want to see this broken down by type of resource (e.g. compute, storage, specialized hardware like GPUs).

* The _Notebook Servers_ view is a decent place to show this information.
* It's okay if the reported costs include all resources of the selected namespace. It doesn't have to be limited to just the resources displayed in the current view (e.g. `notebook` resource type).

saffaalvi:
@sylus 
<img width="968" alt="Screen Shot 2020-10-13 at 4 24 04 PM" src="https://user-images.githubusercontent.com/57377830/95912212-842d6d80-0d70-11eb-8a50-5a11a1695634.png">

sylus:
Wow @saffaalvi that is so incredibly cool, great job!

saffaalvi:
@sylus Thank you!

PR #27, opened 2020-10-06
feat: add cost estimate table

saffaalvi:
Resolves https://github.com/StatCan/jupyter-apis/issues/25


frontend/src/app/main-table/cost-table/cost-table.component.ts
```diff
@@ -0,0 +1,58 @@
...
+  poller: ExponentialBackoff;
+
+  constructor(
+    private namespaceService: NamespaceService,
+    private kubecostService: KubecostService,
+  ) { }
+
+  ngOnInit() {
+    // Create the exponential backoff poller
+    this.poller = new ExponentialBackoff({ interval: 2000, retries: 3 });
```
brendangadd, line 29: This doesn't seem to be used.

frontend/src/app/main-table/resource-table/resource-table.component.ts
```diff
@@ -26,14 +41,55 @@ export class ResourceTableComponent implements OnInit {
...
+    private k8s: KubernetesService,
     private dialog: MatDialog
-  ) {}
+  ) { }
+
+  ngOnInit() {
+    this.dataSource.sort = this.sort;
+
+    // Create the exponential backoff poller
+    this.poller = new ExponentialBackoff({ interval: 2000, retries: 3 });
```
brendangadd, line 56: This component isn't supposed to fetch data. Should adding the cost table be modifying the `resource-table` at all?

namespacecost.go
```diff
@@ -0,0 +1,43 @@
...
+func (s *server) GetCost(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	namespace := vars["namespace"]
+
+	log.Printf("loading namespace costs for %q", namespace)
+
+	params := url.Values{}
+	params.Add("namespace", namespace)
+
+	resp, err := http.Get("http://localhost:9090/model/aggregatedCostModel?window=1d&aggregation=namespace&" + params.Encode())
```
brendangadd, line 22: * URL needs to be an application parameter. This will only work locally.
* Use `url.Values` for either all query parameters, or none of them (via `QueryEscape`).

namespacecost.go
```diff
@@ -0,0 +1,43 @@
...
+	if err != nil {
+		s.error(w, r, err)
+		return
+	}
+
+	responseString := string(body)
+	var response interface{}
+	json.Unmarshal([]byte(responseString), &response)
+
+	s.respond(w, r, response)
```
brendangadd, line 41: Don't use `respond` here as it doesn't do what you need. You're really just passing along the response from the Kubecost API, so just write out the status, content type, and body directly from the response.

frontend/src/app/main-table/cost-table/cost-table.component.ts
```diff
@@ -0,0 +1,58 @@
...
+    this.poller.reset();
+    this.getAggregatedCost();
+  }
+
+  formatCost(value: number): string {
+    return '$' + (value > 0 ? Math.max(value, 0.01) : 0).toFixed(2)
+  }
+
+  getAggregatedCost() {
+    this.kubecostService.getAggregateCost(this.currNamespace).subscribe(
```
brendangadd, line 54: Don't make this component responsible for fetching data. Instead, follow the pattern used in the main table component and existing children.

brendangadd (changes requested):

namespacecost.go
```diff
@@ -0,0 +1,35 @@
...
+	namespace := vars["namespace"]
+
+	log.Printf("loading namespace costs for %q", namespace)
+
+	params := url.Values{}
+	params.Add("namespace", namespace)
+	params.Add("window", "1d")
+	params.Add("aggregation", "namespace")
+
+	resp, err := http.Get(kubecostURL + "/model/aggregatedCostModel?" + params.Encode())
```
zachomedia, line 22:
```suggestion
	u, err := url.Parse(kubecostURL)
	if err != nil {
		s.error(w, r, err)
		return
	}
	
	u, err = u.Parse("model/aggregatedCostModel")
	if err != nil {
		s.error(w, r, err)
		return
	}
	
	params := u.Query()
	params.Add("namespace", namespace)
	params.Add("window", "1d")
	params.Add("aggregation", "namespace")
	u.RawQuery = params.Encode()

	resp, err := http.Get(u.String())
```

I have a preference for using the built-in objects for URL parsing then just string-appending. @brendangadd thoughts?

brendangadd: @zachomedia Sure thing. In that case, I recommend:
* `url.Parse` the Kubecost URL at startup so we can:
  - Fail fast with any resulting `err`
  - Pass the URL object to dependent components rather than a string
  - Avoid parsing on every request
* Suppress `err` from the subsequent `Parse` for path concatenation; I don't think it's possible for that constant path to generate an error.
* Maintain the `params := url.Values{}` approach as `u.Parse("some/path")` will clobber any existing query (i.e. `u.Query()` will always be an empty `url.Values`).

/cc @saffaalvi

brendangadd (approved):
Glorious!

Issue #33, opened 2020-10-27
fix error message when running locally


saffaalvi:
When this component is ran locally, it can't connect to the Kubecost API which causes it to crash with the error:

<img width="915" alt="Screen Shot 2020-10-27 at 2 25 49 PM" src="https://user-images.githubusercontent.com/57377830/97368152-b36bd080-1880-11eb-89ff-45b2a28a7643.png">

Can be fixed by implementing a check for an empty response from the API when populating the table with cost values.

PR #34, opened 2020-10-27
fix: add check for empty response

saffaalvi:
Resolves https://github.com/StatCan/jupyter-apis/issues/33

- Checks to see if response from API is not empty
- Fix cost table message to be more descriptive (over the last 24 hours instead of per day) as it is confusing to some users

### DIFF
--- a/frontend/src/app/main-table/cost-table/cost-table.component.html
+++ b/frontend/src/app/main-table/cost-table/cost-table.component.html
@@ -9,7 +9,10 @@
   </p>
 
   <mat-divider></mat-divider>
-  <table style="width: auto">
+  <table
+    *ngIf="getStatus() == AsyncStatus.SUCCESS; else error"
+    style="width: auto"
+  >
     <tr class="mat-header-row" style="text-align: left">
       <th class="mat-header-cell">{{ "costTable.thCompute" | translate }}</th>
       <th class="mat-header-cell">{{ "costTable.thGpus" | translate }}</th>
@@ -18,13 +21,23 @@
         {{ "costTable.thTotal" | translate }}
       </th>
     </tr>
-    <tr *ngIf="aggregatedCost?.data[currNamespace]; let cost" class="mat-row" style="text-align: left;">
+    <tr
+      *ngIf="aggregatedCost.data[currNamespace]; let cost"
+      class="mat-row"
+      style="text-align: left"
+    >
       <td class="mat-cell">{{ formatCost(cost.cpuCost + cost.ramCost) }}</td>
       <td class="mat-cell">{{ formatCost(cost.gpuCost) }}</td>
       <td class="mat-cell">{{ formatCost(cost.pvCost) }}</td>
-      <td class="mat-cell" style="font-weight: 500;">
+      <td class="mat-cell" style="font-weight: 500">
         {{ formatCost(cost.totalCost) }}
       </td>
     </tr>
   </table>
+
+  <ng-template #error>
+    <div *ngIf="getStatus() == AsyncStatus.FAILURE" class="costerror">
+      {{ "costTable.errMessage" | translate }}
+    </div>
+  </ng-template>
 </div>
--- a/frontend/src/app/main-table/cost-table/cost-table.component.scss
+++ b/frontend/src/app/main-table/cost-table/cost-table.component.scss
@@ -3,13 +3,21 @@ p {
   padding-right: 16px;
 }
 
+.costerror {
+  padding: 20px;
+  padding-left: 24px;
+  color: #cc0033;
+}
+
 .header p {
   padding: 12px 0 0px;
   font-weight: 400;
   font-size: 20px;
 }
 
-.mat-cell, .mat-header-cell, .mat-footer-cell {
+.mat-cell,
+.mat-header-cell,
+.mat-footer-cell {
   padding-left: 12px;
   padding-right: 195px;
 }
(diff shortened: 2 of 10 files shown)


frontend/src/app/main-table/cost-table/cost-table.component.html
```diff
@@ -5,23 +5,29 @@
...
 
   <p>
-    Estimated cost per day of notebook server resources – summed values may not
+    Estimated cost of notebook server resources over the last 24 hours – summed values may not
     match total due to rounding
   </p>
 
   <mat-divider></mat-divider>
-  <table style="width: auto">
+  <table *ngIf="aggregatedCost!==null && aggregatedCost.data!==undefined; else serverError" style="width: auto">
```
brendangadd, line 13: Looks like `aggregatedCost` is `null` until a successful response occurs. So this would result in a failure message to the user while waiting for a response (which may succeed).

frontend/src/app/main-table/cost-table/cost-table.component.html
```diff
@@ -5,23 +5,29 @@
   </div>
 
   <p>
-    Estimated cost per day of notebook server resources – summed values may not
+    Estimated cost of notebook server resources over the last 24 hours – summed values may not
```
brendangadd, line 8: What's the reason for this change?

brendangadd (changes requested):

frontend/src/app/main-table/cost-table/cost-table.component.html
saffaalvi: @ca-scribner and I felt it was confusing to some users as it implied it was the cost of resources for the current day, so if a new resource was added, user might think the cost should update.

ca-scribner: yeah the estimate given here isn't a forward-looking per-day cost estimate (I don't think?), but instead a backward 24-hour assessment of what you used.  So if today I run nothing, then tomorrow I turn on a GPU and lots of expensive stuff and immediately look at kubecost, it'll still say "Estimated cost per day of notebook server resources = $0".  But tomorrow if I look again, it'll be $$$.  So the wording change was suggested to make it clear we're estimating based on history, not current running resources

Even better would be if we did our own forward-looking estimate based on a shorter timeframe if that's possible.  for example, do something like "last minute of costs * 24 * 60 * 60 = estimated cost per day based on your usage in the last minute"

...unless I misunderstood kubecost's output.  If it is forward looking already, then the old wording is great

brendangadd: :+1:  `Estimated -> Approximate`

frontend/src/app/services/kubecost.service.ts
```diff
@@ -48,18 +49,24 @@ export class KubecostService {
...
   }
 
-  private handleError(
-    error: HttpErrorResponse | AggregateCostResponse
-  ): Observable<never> {
+  private handleError(error: HttpErrorResponse | AggregateCostResponse | Resp): Observable<never> {
+    if (this.isResp(error)) {
+      return throwError(new Error(error.log));
+    }
     return throwError(new Error(error.message));
```
brendangadd:
```suggestion
    const message = this.isResp(error) ? error.log : error.message;
    return throwError(new Error(message));
```

frontend/src/app/services/kubecost.service.ts
```diff
@@ -48,18 +49,24 @@ export class KubecostService {
...
         tap(res => this.handleBackendError(res)),
         catchError(err => this.handleError(err))
-      );
+      ) as Observable<AggregateCostResponse>;
   }
 
-  private handleBackendError(response: AggregateCostResponse) {
-    if (response.code < 200 || response.code >= 300) {
+  private handleBackendError(response: AggregateCostResponse | Resp) {
+    if (this.isResp(response) || response.code < 200 || response.code >= 300) {
```
brendangadd, line 56: This is nice.

frontend/src/app/main-table/cost-table/cost-table.component.ts
```diff
@@ -9,27 +15,22 @@ import { AggregateCostResponse } from 'src/app/services/kubecost.service';
...
-      if (Object.keys(this.aggregatedCost).includes('log')) {
-        this.message = "Failed to retrieve cost information."
-        return false;
-      }
-      else {
-        return true;
-      }
-    }
+  getStatus(): AsyncStatus {
+    if (this.currNamespace=='' || this.aggregatedCost==null) {
```
brendangadd, line 26: I don't think this component should know about the current namespace. That is a concern of the parent component. If the parent wants to reset the state of this table when the namespace changes, then it can set the `aggregatedCost` parameter to `null`.

frontend/src/app/main-table/cost-table/cost-table.component.scss
```diff
@@ -2,6 +2,15 @@ p {
   padding-left: 24px;
   padding-right: 16px;
 }
+.waiting {
+  padding: 20px;
+  padding-left: 24px;
+}
```
brendangadd, line 8: Is this used?

brendangadd (changes requested):
Your changes need formatting. @wg102 can show you how to make Prettier only format the parts you've changed when you save in VS Code, which will avoid adding ~20 files worth of formatting to this PR. :)

frontend/src/app/main-table/cost-table/cost-table.component.scss
saffaalvi: not anymore - removed!

brendangadd (dismissed):

Issue #37, opened 2020-11-03
Enforce that new GPU notebook servers requests book the entire machine


ca-scribner:
GPU nodes are used solely for workflows which need GPUs due to ~10x costs per node (GPU node = $3-4/hr, regular node ~0.46/hr @ 1-year reserved rate, 0.768/hr if not reserved).  Costing passed to users is prorated by how much of a machine a user requests (request half a machine's CPU/RAM, pay for half the machine's costs).  

Any user that requests a GPU will result in a new full GPU node spun up.  Regardless of how much of the node the user occupies, the platform is charged for a full node.  But if the user reserves only a part of that node (say 20% of RAM and 20% of CPU, plus the GPU) they don't get billed for the rest of the idle CPU/RAM they've caused.

Suggested solution: require that a GPU notebook server be booked as a full machine, therefore passing the full cost to the user.  In our case, that means requiring that any GPU request be for 5CPU/96GB RAM.

ca-scribner:
This could be completed alongside #36

ca-scribner:
(before anyone actually does this, we need agreement on if we should do it.  I think we should, but that wasn't a concensus)

Nancy-Badran:
ok, good to know

blairdrummond:
@ca-scribner I think this makes sense. Also, confirming what @chritter thought, the input from the notebook-server form fills the `require` and `limit` fields

```
Containers:                    
  tf-test:                               
    Image:      k8scc01covidacr.azurecr.io/blairs-vscode-tf
    Port:       8888/TCP                                                                                                                                                                       
    Host Port:  0/TCP             
    Limits:      
      cpu:             1                                                                                                                                                                       
      memory:          2Gi       
      nvidia.com/gpu:  1
    Requests:       
      cpu:             1                                                                       
      memory:          2Gi
      nvidia.com/gpu:  1
    Environment:          
      NB_PREFIX:  /notebook/blair-drummond/tf-test
```

So jobs would get evicted (unnecessarily) if they use more than 2gb of RAM on those GPU nodes.

So setting the fields to high/max values also will ensure that no jobs get evicted unnecessarily.

PR #45, opened 2020-11-24
fix: add cpu/ram max requirement for 1 gpu (frontend) + translation

saffaalvi:
Resolves https://github.com/StatCan/jupyter-apis/issues/37

Wasn't sure if we want it as an error if the user doesn't enter the max CPU/RAM numbers if 1 GPU is selected or if we want to grey out the input areas for CPU/RAM and set it as a default? 
For now, did it as part of the error check so the user is aware they are booking the entire machine.

### DIFF
--- a/frontend/src/app/resource-form/resource-form.component.ts
+++ b/frontend/src/app/resource-form/resource-form.component.ts
@@ -27,6 +27,8 @@ export class ResourceFormComponent implements OnInit, OnDestroy {
 
   subscriptions = new Subscription();
 
+  readonlySpecs: boolean;
+
   constructor(
     private namespaceService: NamespaceService,
     private k8s: KubernetesService,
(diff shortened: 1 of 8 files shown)


frontend/src/app/resource-form/resource-form.component.ts
```diff
@@ -103,4 +105,15 @@ export class ResourceFormComponent implements OnInit, OnDestroy {
...
   }
+  
+  // Automatically set values of CPU and Memory if GPU is 1
+  checkGPU(gpu: string) {
+    if (gpu == "none") {
+      this.readonlySpecs = false;
+    } else {
+      this.readonlySpecs = true;
+      this.formCtrl.get("cpu").setValue("5");
+      this.formCtrl.get("memory").setValue("96Gi");
```
brendangadd, line 116: Same question here re. imperative form here vs. declarative via template. Though it makes more sense to me in this case.

brendangadd (changes requested):

brendangadd (approved):

PR #46, opened 2020-12-16
feat: Add 1TB/2TB drive for notebook servers

ca-scribner:
Add 1TB and 2TB drive options as per DScD customer request


saffaalvi (approved):

brendangadd (changes requested):
.

Issue #47, opened 2020-12-17
Pass KF_LANG to the Notebook Server


wg102:
Add an environment variable called KF_LANG to the pod. Defaulting it's value to the browser language. 
In the "New server", the user can override it to select either French or English. 
This should be under Advanced Options, which could be a collapsible section.

wg102:
This can be tested by running the code locally, while being on the cluster, and using the custom image k8scc01covidacr.azurecr.io/rstudio:09c968225ec2ef9ced718b3b64567049e0f53a74

PR #48, opened 2020-12-17
feat: add KF_LANG env variable

wg102:
Resolves https://github.com/StatCan/jupyter-apis/issues/47

### DIFF
--- a/frontend/src/assets/i18n/fr.json
+++ b/frontend/src/assets/i18n/fr.json
@@ -48,9 +48,11 @@
     "deleteDialogNo": "Annuler"
   },
   "formAdvancedOptions": {
-    "h3MiscSettings": "Paramètres divers",
-    "txtMiscSettings": "Autres paramètres pouvant être appliqué au serveur bloc-notes.",
-    "toggleSharedMemory": "Activer la mémoire partagée"
+    "h3AdvancedSettings": "Paramètres avancés",
+    "toggleSharedMemory": "Activer la mémoire partagée",
+    "lblSystemLanguage":"Langue du système",
+    "lblEnglish": "English",
+    "lblFrench": "Français"
   },
   "formConfiguration": {
     "h3Configuration": "Configurations",
--- a/notebooks.go
+++ b/notebooks.go
@@ -3,10 +3,12 @@ package main
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"log"
 	"net/http"
+	"regexp"
 	"sort"
 	"strings"
 
@@ -23,6 +25,7 @@ import (
 const DefaultServiceAccountName string = "default-editor"
 const SharedMemoryVolumeName string = "dshm"
 const SharedMemoryVolumePath string = "/dev/shm"
+const EnvKfLanguage string = "KF_LANG"
 
 type volumetype string
 
@@ -61,6 +64,7 @@ type newnotebookrequest struct {
 	DataVolumes        []volumerequest   `json:"datavols"`
 	EnableSharedMemory bool              `json:"shm"`
 	Configurations     []string          `json:"configurations"`
+	Language           string            `json:"language"`
 }
 
 type notebookresponse struct {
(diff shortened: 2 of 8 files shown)


Ito-Matsuda (changes requested):
Unsure if the following files also need to be modified to also have a field for language; [types.ts](https://github.com/StatCan/jupyter-apis/blob/feat-env-var/frontend/src/app/utils/types.ts) and [spawner_ui_config.yaml](https://github.com/StatCan/jupyter-apis/blob/feat-env-var/samples/spawner_ui_config.yaml). Only found those two files after searching with `shm` through this project.  The `spawner_ui_config` seems to set defaults and if it's read only, but the code works just fine without it (set elsewhere). As for `types.ts` I have a feeling it _might_ (make note of italics im not 100% sure) need to be in there somewhere.  I'm not super familiar with jupyter-apis but searching [import { Config } ](https://github.com/StatCan/jupyter-apis/search?q=import+%7B+Config+%7D) produces some results where the `Config` interface definition is used.

notebooks.go
```diff
@@ -61,6 +62,7 @@ type newnotebookrequest struct {
 	DataVolumes        []volumerequest   `json:"datavols"`
 	EnableSharedMemory bool              `json:"shm"`
 	Configurations     []string          `json:"configurations"`
+	Language		   string			 `json:"language"`
```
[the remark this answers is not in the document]
wg102: This was due to using tabs instead of spaces

notebooks.go
```diff
@@ -471,6 +475,19 @@ func (s *server) NewNotebook(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	//Add Language
+	//Validate that the language format is valid (language[_territory])
+	match, err := regexp.MatchString("[[:alpha:]]{2}(_[[:alpha:]]{2})?$", req.Language)
```
brendangadd, line 480: Matches anything that **ends** with a valid code. Needs a `^` in there.

notebooks.go
```diff
@@ -23,6 +25,7 @@ import (
 const DefaultServiceAccountName string = "default-editor"
 const SharedMemoryVolumeName string = "dshm"
 const SharedMemoryVolumePath string = "/dev/shm"
+const EnvKFLanguage string = "KF_LANG"
```
brendangadd, line 28: Treat acronyms in variable names as regular words for camelcasing.

notebooks.go
```diff
@@ -471,6 +475,19 @@ func (s *server) NewNotebook(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	//Add Language
+	//Validate that the language format is valid (language[_territory])
+	match, err := regexp.MatchString("[[:alpha:]]{2}(_[[:alpha:]]{2})?$", req.Language)
+		if (err != nil || !match) {
+			var errLanguageFormat = errors.New("Error server language format")
```
brendangadd, line 482: Make error message more clear, so someone reading logs in a console is likely to understand what occurred.

Ito-Matsuda (commented):
Looks fine to me. Right now `shm` defaults to True/Enabled, but you mentioned that the change there needs to take place in the manifest.

frontend/src/assets/i18n/fr.json
```diff
@@ -48,9 +48,11 @@
     "deleteDialogNo": "Annuler"
   },
   "formAdvancedOptions": {
-    "h3MiscSettings": "Paramètres divers",
-    "txtMiscSettings": "Autres paramètres pouvant être appliqué au serveur bloc-notes.",
-    "toggleSharedMemory": "Activer la mémoire partagée"
+    "h3AdvancedSettings": "Paramètres avancés",
+    "toggleSharedMemory": "Activer la mémoire partagée",
+    "lblSystemLanguage":"Language du système",
```
brendangadd, line 53: "Langue"?

brendangadd (changes requested):
**Commit message:**
* Make title describe the feature rather than the variable
* Add a message body that provides some details on what you changed

brendangadd:
@wg102 Looks great, and commit message is excellent. You can squash it down (keep the new commit message) and we'll merge.

Issue #50, opened 2020-12-17
Validate notebook server PVCs on server creation


ca-scribner:
Current state: 

Users can try to create notebook servers that request mounting a read-write-once (single user) data volume that is already in use.  This results in a notebook server that will never fully start

Proposal:

Validate the user's request for volumes on the new notebook server prior to submission, alerting them to this error.  The page already shows PVC status, so this information should be accessible somehow.  This way users don't get stuck with machines that never start, which is both frustrating and causes them to ask us for help.

wg102:
The first commit to the draft PR adds the error message in the model. But it does not automatically show in a specific case: when the element has yet to be touched. Currently working on this, the solution should be something specific to material angular.

PR #51, opened 2021-01-04
feat: Add validation for mounted pvc

wg102:
Resolves https://github.com/StatCan/jupyter-apis/issues/50


ca-scribner:
This is a nice to have but not urgent.  Suggest we wait for Saffa to return and get her to review rather than Brendan/Will?

saffaalvi (approved):

PR #61, opened 2021-05-19
fix: add url validation for custom images

### LINKED ISSUE #59: Add validation to image_name in notebook server creation to prevent or strip leading "http[s]://"
When creating a notebook server you can enter an image name that has a leading `http[s]://`, but this will never work (the name used here is a docker image, eg: `[repo/]imagename:tag`).  The frontend can validate this input and do one of:
* strip leading `http[s]://` from any image name without the user knowing it had to be done (preferable unless there's a use case that ever needs the leading http)
* warn the user when filling out he form (or reject outright) that http is not accepted (do this if we don't think we can just safely always strip)

jumana-s:
Resolves #59

### DIFF
--- a/frontend/src/app/resource-form/form-image/form-image.component.ts
+++ b/frontend/src/app/resource-form/form-image/form-image.component.ts
@@ -1,5 +1,7 @@
-import { Component, OnInit, Input } from "@angular/core";
-import { FormGroup } from "@angular/forms";
+import {Component, OnInit, Input} from "@angular/core";
+import {FormGroup, ValidatorFn, AbstractControl} from "@angular/forms";
+import {TranslateService} from "@ngx-translate/core";
+import {Subscription} from "rxjs";
 
 @Component({
   selector: "app-form-image",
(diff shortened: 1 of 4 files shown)


frontend/src/app/resource-form/form-image/form-image.component.ts
```diff
@@ -12,24 +19,66 @@ export class FormImageComponent implements OnInit {
...
+      return this.translate.instant("formImage.errorHttp", {
+        http: `${urlBeginning}`
+      });
+    }
+  }
+
+  private urlValidator(): ValidatorFn {
+    return (control: AbstractControl): {[key: string]: any} => {
+      const valid = control.value.search(/^http[s]?:\/\//i);
+      return valid != -1 ? {validUrl: true} : null;
```
wg102, line 81: Why not use RegExp directly instead? (bis)

frontend/src/app/resource-form/form-image/form-image.component.ts
```diff
@@ -12,24 +19,66 @@ export class FormImageComponent implements OnInit {
...
     return displayName;
   }
+
+  urlValidation(): void {
+    const url = this.parentForm.get("customImage");
+
+    if (url.hasError("validUrl")) {
+      let urlBeginning;
+
+      if (url.value.search(/http:\/\//i) != -1) {
```
wg102, line 66: Why not use RegExp directly?

wg102 (changes requested):
DS_Store file is something specific to MacOS, and should not be added to the project.

wg102 (approved):

PR #99, opened 2022-03-07
fix(frontend): delete volume tooltip

wg102:
Small fix to go with


Jose-Matsuda (changes requested):
I get 
image

Which is an improvement over what we have on dev but i think not intentional

Jose-Matsuda (approved):
LGTM, this is more than what we currently have on Prod (as in a tooltip) and is apparently difficult to get the 'name' displayed in the tooltip as well so this was the path of least resistance while getting something out

PR #105, opened 2022-03-08
fix: missing key in status of notebook

wg102:
Fix status of both Notebook and Volume statuses


Jose-Matsuda (approved):
Works on my end can see the tooltips for both the notebook status and the notebook volume status

PR #154, opened 2022-10-04
add: NB_Type env var

rohank07:


vexingly:
This should work fine, but we need to make sure we never use this to approve/apply protb related services, deny only type of logic... since someone can easily change their env variables!

vexingly (approved):

rohank07:
On second thought let me try the downwardAPI Volume :smile:

Issue #158, opened 2022-10-05
1.6: Fix volumes


Jose-Matsuda:
This can be a few things

1)We brought in our changes, but there is a failure upon trying to create a notebook server

The `POST` is failing 

![image](https://user-images.githubusercontent.com/23174198/194075152-57c68be8-7574-4af2-9356-631daeb40ef5.png)

Which is found and seen [here](https://github.com/StatCan/jupyter-apis/blob/9fcef881dd8eeda371ded319aa5c4dee00812f63/notebooks.go#L388), we are sending an empty string as a request for the volume type.

2) We might need to send something / fix something up for storageclasses(#159 ), I know @mathis-marcotte will be creating the endpoint for that but there might be another thing not too sure

Yeah as Mathis probably noted the request format changed

mathis-marcotte:
This error is also probably related, which appears when you first open the form
![image](https://user-images.githubusercontent.com/84033116/194076940-98ec4a91-431d-4597-8272-db7bbd1280e8.png)

Jose-Matsuda:
## How it was done previously

I'm really not sure how it does work, maybe it does some stuff [here](https://github.com/StatCan/jupyter-apis/blob/3f6d99be74258e160f1d7c0397630bfd10b5a9a8/frontend/jupyter/src/app/pages/form/form-default/volume/volume.component.html#L6) need to look more into things.

Jose-Matsuda:
## List of Files that must change if we move to upstream's way

1) `notebooks.go`, this is where we get the error above, and in here we have definition of the `struct`s as well as the `unmarshal`s which populate the data that we retrieve from the frontend requests.

2) `config.go`, I believe this, in tandem with `spawner_ui_config.yaml` is the cause for the error mentioned by Mathis, there are again definitions of the volumes and expected request / responses.

Jose-Matsuda:
Ok I tested in the following [gist](https://gist.github.com/Jose-Matsuda/4caa1bf34348f322a71a13825086867a) what happens if I created a struct that may or may not hold empty values, turns out its just ok (which is nice) as we might be able to just have a single struct which will hold the possible cases for each request.

Jose-Matsuda:
Initial attempt from [this commit](https://github.com/StatCan/jupyter-apis/commit/811068a2f55ada23ead26e5b086240c7bee59408)

[...]

PR #164, opened 2022-10-12
1.6 fix volumes

Jose-Matsuda:
Fix volumes let people submit their workload requests

closes https://github.com/StatCan/jupyter-apis/issues/158

### DIFF
--- a/config.go
+++ b/config.go
@@ -64,6 +64,7 @@ type ValueElement struct {
 }
 
 type ValueValue struct {
+	//this has to change
 	Type        ImagePullPolicy `yaml:"type" json:"type"`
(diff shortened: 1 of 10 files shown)


mathis-marcotte:
With the data volumes, I've noticed a few things. So when you add a data volume, firstly with the name of the data volume the name gets generated as "[object Object]-datavol-1" instead of using the name of the notebook. Second, and this one is happening on both data and workspace volumes, in that info line next to the "New Volume" it defaults with the value 16Gi, but when you open the volume information, that "Size in Gi" input is empty. Lastly, when I add a data volume, even if I fill things in the form, the launch button at the end stays disabled. 
image

mathis-marcotte (approved):

PR #174, opened 2022-10-20
fix(volume name): update formCtrl.get()

bryanpaget:
fix(volume name): update formCtrl.get() to formCtrl.get().value.

The Angular code for `<app-form-data-volumes>` was missing `.value`, I just needed to add it. :-)


bryanpaget:
The same issue appears in the upstream code:

https://github.com/kubeflow/kubeflow/blob/538c2ad6d973072783db5f61a374ef8d8d6030c1/components/crud-web-apps/jupyter/frontend/src/app/pages/form/form-default/form-default.component.html#L55

PR #175, opened 2022-10-21
set(sizeNum): default to 16

### LINKED ISSUE #108: JWA: Fix default dropdown for Workspace Volume
Currently looks like 
image

It will work and use the default 16Gi from the spawner_config if created without making a selection but would rather have it start on `16`.

There has been some initial work since last sprint trying to get this to work via fiddling around with the various `ng` lifecycle statements but everytime the value has been overwritten.

Probably same issue for data volumes?
image

bryanpaget:
closes #108 

### Description:
Setting `this.sizeNum.setValue("16");` in `ngOnInit()` of `size.component.ts` sets the default value in `mat-select` of `size.component.html` to `16`.

### Screenshot:
image

### Reference:
- https://stackoverflow.com/questions/50650790/set-default-option-in-mat-select


Jose-Matsuda (approved):
LGTM. Everything worked well during my testing. Created a workload with a data and a workspace volume, both had the default setting of 16Gi.

Created two files in the two paths, deleted the workspace, and then created a new workspace mounting both those volumes and opened it up and voila everything still worked fine

Issue #178, opened 2022-10-26
UPSTREAM: Revert the CPU and RAM to default settings when toggling the number of GPUs.


bryanpaget:
## Introduction
It's not hard to have the CPU and Memory settings revert back to whatever they were before enabling a GPU. The issue is do we want to maintain this change? I'll try to get upstream to accept the change.

### Description
In `form-default.component.ts` we have a method to check the GPU configuration (this is what it currently looks like):
```
  // Automatically set values of CPU and Memory if GPU is 1
  checkGPU(gpu: string) {
    if (gpu == "none") {
      this.readonlySpecs = false;
    } else {
      this.readonlySpecs = true;
      this.formCtrl.get("cpu").setValue("4");
      this.formCtrl.get("memory").setValue("96");
    }
  }
```
We could add something like:
```
  // Automatically set values of CPU and Memory if GPU is 1
  checkGPU(gpu: string) {
    if (gpu == "none") {
      this.readonlySpecs = false;
      this.formCtrl.get("cpu").setDefaultValue();
      this.formCtrl.get("memory").setDefaultValue();
    } else {
      this.readonlySpecs = true;
      this.formCtrl.get("cpu").setValue("4");
      this.formCtrl.get("memory").setValue("96");
    }
  }
```
Where `.setDefaultValue()` is a made up method.

bryanpaget:
It turns out this code is not available upstream at all:



![Image](https://user-images.githubusercontent.com/8212170/218528235-846f7017-5440-47fd-988f-ec7156de72c0.png)

It's only implemented in our fork:

![Image](https://user-images.githubusercontent.com/8212170/218528149-87d9ceed-8ea9-4790-8efe-7f5d114776ee.png)

bryanpaget:
I think the new code for setting the notebook server configuration is found here:

https://github.com/kubeflow/kubeflow/blob/6b899b2ad9848fe3051b2e430bd6785866b9fdda/components/crud-web-apps/jupyter/backend/apps/common/form.py



![Image](https://user-images.githubusercontent.com/8212170/218574074-e6c0ed54-e298-47e1-a038-a640b849d47d.png)

bryanpaget:
So far I have the following proposal but I would prefer to access the default CPU and Memory values stored in `spawner_ui_config.yaml`, sticking to a central place for configuration.

```
  // Automatically set values of CPU and Memory if GPU is 1
  checkGPU(gpu: string) {    
    if (gpu == "none") {
[...]

PR #202, opened 2023-02-16
update(form-default): return to usable state.

bryanpaget:
When the user toggles the GPU settings, the CPU and memory settings now returns to 1 CPU and 4 gigabyes of ram, which is a usable state.


bryanpaget:
Closes #178.

Issue #221, opened 2023-04-17
Update jupyter-apis i18n to KF1.7


mathis-marcotte:
Once all the code is updated for the frontend, we should make sure to run `npm run i18n:extract` to update the messages files for our translations, and also add any missing translation.

Since updating xliff files can be annoying, maybe we can look into a tool like [xliffmerge](https://www.npmjs.com/package/ngx-i18nsupport) to help us out.

List of files 
# I18n
jupyter/frontend/i18n/fr
- [x] OWNERS
- [x] messages.fr.xlf

jupyter/frontend/i18n/
- [x] messages.xlf

jupyter/frontend/src/app/pages/form/form-new/form-cpu-ram
- [x] form-cpu-ram.component.html

jupyter/frontend/src/app/pages/form/form-new/form-data-volumes
- [x] form-data-volumes.component.html

jupyter/frontend/src/app/pages/form/form-new/form-image
- [x] form-image.component.html
There is also new text we will need to figure out for this one for the description of the SAS image

wg102:
# Notes from other tickets: 

projects/kubeflow/src/lib/confirm-dialog:
- types.ts : Will need to translate?
-
projects/kubeflow/src/lib/content-list-item:
- Entirely new thing. Will need to be translated

projects/kubeflow/src/lib/editor: New folder
- Will need translation?

projects/kubeflow/src/lib/editor/interfaces: New folder
- Will need translation?

projects/kubeflow/src/lib/form/rok-url-input:
- New content but this is all untranslated. We don't use rok.


section.component.html : Customization conflict.
- New help section
projects/kubeflow/src/lib/form/section-image:
- Will need to investigate when in that section and make the necessary adjustments.

projects/kubeflow/src/lib/help-popover: New

- New

projects/kubeflow/src/lib/loading-spinner: 
- new css class may needed translation

projects/kubeflow/src/lib/logs-viewer: New



namespace-select.component.html 
- Will need to confirm new option
namespace-select.component.ts 
- New input + lots of code changes

projects/kubeflow/src/lib/resource-table/action:
The following might be translation troubles:
- action.component.html went from matTooltip="{{ tooltipReady }}" to matTooltip="{{ action.tooltip }}"
- action.component.html went from matTooltip="{{ tooltipInit }}" to matTooltip="{{ action.tooltip }}"

projects/kubeflow/src/lib/resource-table/component-value:
- component-value.component.ts New output

projects/kubeflow/src/lib/resource-table/types:
[...]

Issue #218, opened 2023-04-17
Update jupyter-apis index frontend to KF1.7


mathis-marcotte:
Epic https://github.com/StatCan/aaw/issues/1632

Update the frontend for the index section of jupyter-apis to match [upstream](https://github.com/kubeflow/kubeflow/tree/v1.7.0/components/crud-web-apps/jupyter/frontend/src/app/pages/index) while making sure to not remove our custom code.

~~For the volume table, make sure to look in [here](https://github.com/kubeflow/kubeflow/tree/v1.7.0/components/crud-web-apps/volumes).~~ This will be another ticket

~~This would also include adding the [notebook details](https://github.com/kubeflow/kubeflow/tree/v1.7.0/components/crud-web-apps/jupyter/frontend/src/app/pages/notebook-page) and the [volume details](https://github.com/kubeflow/kubeflow/tree/v1.7.0/components/crud-web-apps/volumes/frontend/src/app/pages/volume-details-page) pages.~~ another ticket

## SRC App Pages Index
jupyter/frontend/src/app/pages/index
- [x] index.component.html
- [x] index.component.scss
- [x] index.component.spec.ts
- [x] index.component.ts
- [x] index.module.ts

jupyter/frontend/src/app/pages/index/index-default
- [x] config.ts : Seems like more properties got added to the table column configs, mainly the `sort`. We'll have to make sure to update our column definitions if needed for the volume and cost tables.
- [x] index-default.module.ts
- [x] index-default.component.scss
- [x] index-default.component.html
- [x] index-default.component.spec.ts
- [x] index-default.component.ts : The way the backend is called for the notebooks changed. So that caused a bit more conflicts with our other custom calls for the volume and kubecost data.

jupyter/frontend/src/app/pages/index/index-default/server-type
- [x] server-type.component.ts
- [x] server-type.component.scss
- [x] server-type.component.spec.ts
- [x] server-type.component.html

jupyter/frontend/src/app/pages/index/index-rok
(We dont use this component so just copying it all)
- [x] index-rok.component.spec.ts
- [x] index-rok.component.ts
- [x] index-rok.module.ts

PR #235, opened 2023-05-26
Kf1.7 jupyter apis index frontend

mathis-marcotte:
for https://github.com/StatCan/jupyter-apis/issues/218

Added the file index-default.component.ts in the [i18n ticket](https://github.com/StatCan/jupyter-apis/issues/221) since there are some new text in there

### DIFF
--- a/frontend/jupyter/src/app/pages/index/index-default/config.ts
+++ b/frontend/jupyter/src/app/pages/index/index-default/config.ts
@@ -55,9 +55,11 @@ export const defaultConfig: TableConfig = {
       sortingPreprocessorFn: element => element.serverType,
       filteringPreprocessorFn: element => {
         if (element.serverType === 'group-one') {
-          return 'vscode Visual Studio Code';
-        } else if (element.serverType === 'group-two') {
           return 'rstudio';
+        } else if (element.serverType === 'group-two') {
+          return 'ubuntu';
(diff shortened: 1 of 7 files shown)


frontend/jupyter/src/app/pages/index/index-default/config.ts
```diff
@@ -5,87 +5,78 @@ import {
...
-      matHeaderCellDef: $localize`Age`,
+      sort: true,
+      sortingPreprocessorFn: element => element.serverType,
+      filteringPreprocessorFn: element => {
+        if (element.serverType === 'group-one') {
+          return 'vscode Visual Studio Code';
+        } else if (element.serverType === 'group-two') {
+          return 'rstudio';
+        } else {
+          return 'jupyterlab';
```
wg102, line 62: If they mentioned the two groups  + jupyterlab, do we want to add our third group to this option or wait...

wg102 (changes requested):
Questions: 
1. Are Dialogs still working? (They got removed or moved, but I think they still get called)
2. If the answer to the comment in the code is 'we don't care' then it's LGTM

frontend/jupyter/src/app/pages/index/index-default/config.ts
mathis-marcotte: oh you are correct! Ill go update that now

PR #264, opened 2023-07-24
feat(ui) refactor notebook menu

wg102:
Closes https://github.com/StatCan/aaw/issues/1774


Jose-Matsuda (commented):
everything seems to work fine, will let Mathis take his hand at it (using c56241ce5f73dbb2c875585768c08186cd81122d on dev)

frontend/jupyter/yarn.lock
```diff
@@ -6269,12 +6269,6 @@
   "resolved" "https://registry.npmjs.org/klona/-/klona-2.0.6.tgz"
   "version" "2.0.6"
 
-"kubeflow@file:../common/kubeflow-common-lib/dist/kubeflow":
```
mathis-marcotte, line 6272: I don't know how much this actually would affect us, but it's probably best to NOT commit this change.

notebooks.go
```diff
@@ -533,6 +534,11 @@ func (s *server) NewNotebook(w http.ResponseWriter, r *http.Request) {
 		notebook.Spec.Template.Spec.Containers[0].Resources.Limits[corev1.ResourceMemory] = req.MemoryLimit
 	}
 
+	// There's more than just the json
```
mathis-marcotte, line 537: What do you mean that there's more than just the json?

frontend/jupyter/src/app/pages/form/form-new/form-new.component.html
```diff
@@ -29,41 +29,67 @@
...
-      >
-      </app-form-workspace-volume>
-
-      <app-form-data-volumes
-        [volsArray]="formCtrl.get('datavols')"
-        [readonly]="config?.dataVolumes?.readOnly"
-        [externalName]="formCtrl.get('name').value"
-      >
-      </app-form-data-volumes>
+      <lib-form-section title="Protected B">
```
mathis-marcotte, line 32: I feel like we could probably convert this into its own component (like with what you did with the form-image-custom)

frontend/jupyter/src/app/pages/form/form-new/form-image-custom/form-image-custom.component.ts
```diff
@@ -0,0 +1,210 @@
...
+import { V1Namespace } from '@kubernetes/client-node';
+import { Config } from 'src/app/types';
+import { MatCheckboxChange } from '@angular/material/checkbox';
+
+@Component({
+  selector: 'app-form-image-custom',
+  templateUrl: './form-image-custom.component.html',
+  styleUrls: ['./form-image-custom.component.scss'],
+})
+export class FormImageCustomComponent implements OnInit, OnDestroy {
```
mathis-marcotte, line 28: I think you should re-evaluate what needs to be in this component. From what I can tell, this was created from a copy paste of the form-image component, and there's a lot of lingering code from that component that is not needed in here.

mathis-marcotte (changes requested):

notebooks.go
wg102: This is the wrong comment. Fixed

mathis-marcotte (approved):
Changes requested were done and things look good to me.

PR #266, opened 2023-07-27
chore(scanning): migrate from azure container scan to trivy

Jose-Matsuda:
Related to https://github.com/StatCan/aaw-private/issues/112

WIP.
I think the duplication of work that is being seen is because the `build.yml` used to be for PR's and would just build on that, but since we changed our `publish.yml` to also account for PRs with the `auto-deploy` tag (which in the majority of cases we want) it looks like a duplication of work which came from this [pr](https://github.com/StatCan/jupyter-apis/commit/51c6c44b92e752a456719e7c8d2878a06bd750bb) where we added the ability to push to dev acr using the tag.

It wouldn't be a duplication of work if the PR doesnt have the `auto-deploy` tag on, it would just `build`. 
So perhaps a good in-between would be to have the `build.yml` workflow run only if the `auto-deploy` tag isn't on it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,6 +10,7 @@ env:
   KUBECTL_VERSION: 1.21.2
   KUSTOMIZE_VERSION: 4.1.3
   REGISTRY_NAME: k8scc01covidacr
+  TRIVY_VERSION: "v0.43.1"
 
 jobs:
   test:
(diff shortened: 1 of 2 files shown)


Jose-Matsuda:
## Summary of Changes

### `build.yml`
Removed the build + scan since that is being handled and done by the `publish.yml` anyways, but kept file to keep the `Taskfile` tests

### `publish.yml`
Moved the push to after the security scan

Jose-Matsuda:
having a wee bit of trouble understanding building then scanning then pushing, we could have it all in the same step to make it easy albeit a bit whacky looking

Jose-Matsuda:
That or we copy contrib containers and really just have a bit more stuff
https://github.com/StatCan/aaw-contrib-containers/blob/8a25f7e1e617cfe5de70c747b8fc5cb31ec00f49/.github/workflows/publish.yml#L57

PR #267, opened 2023-07-28
Remove Protected B from configuration dropdown

wg102:
Closes https://github.com/StatCan/aaw/issues/1789


wg102:
Closing this commit without merging since it is not here that it needs to be solved, but in notebook.go in aaw-kubeflow-profiles-controller

PR #281, opened 2023-09-11
added custom image message

mathis-marcotte:
closes https://github.com/StatCan/aaw/issues/369

### DIFF
--- a/frontend/jupyter/src/app/pages/form/form-new/form-image-custom/form-image-custom.component.html
+++ b/frontend/jupyter/src/app/pages/form/form-new/form-image-custom/form-image-custom.component.html
@@ -104,6 +104,7 @@
         *ngIf="parentForm?.value.customImageCheck"
       >
         <mat-label i18n>Custom Image</mat-label>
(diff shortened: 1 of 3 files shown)


frontend/jupyter/src/app/pages/form/form-new/form-image-custom/form-image-custom.component.html
```diff
@@ -104,6 +104,10 @@
         *ngIf="parentForm?.value.customImageCheck"
       >
         <mat-label i18n>Custom Image</mat-label>
+        <mat-hint i18n>
+          The image given will be assessed by a vulnerability scan before
```
Jose-Matsuda, line 108: I'd probably change the wording here since while I get what you're going for with the vulnerability scan since we do all scan our images, I feel like it's more of a "We can only allow custom images from select registries", maybe something to post in the dev channel just to get feedback on it.

Jose-Matsuda (changes requested):
Just the change to the actual text, the text looks good and fine in practice.

Jose-Matsuda (approved):

Issue #283, opened 2023-09-12
delete button is always available in the volume details page


mathis-marcotte:
In the volume details page, the delete button at the top right of the page is always available.

This contradicts with our logic in the index page to disable the delete button of a volume when that volume is attached to a notebook.

![image](https://github.com/StatCan/jupyter-apis/assets/84033116/f525c30f-c4f0-49fc-8910-c7f0eceae2c6)

PR #291, opened 2023-09-21
Volume details delete button

mathis-marcotte:
closes https://github.com/StatCan/jupyter-apis/issues/283

I had to update cypress version. There was an issue that was failing the github action that runs our cypress tests. Updating the version should fix it.


Jose-Matsuda (approved):
Yup lgtm
image

and then deleting the notebook thus freeing up the volume
image

Issue #257, opened 2023-07-11
Add visual loading indicator for when data takes longer to load


mathis-marcotte:
Sometimes it feels like it takes a little bit of time to load the data in the tables, so it appears for a little bit of time as if there is nothing in the namespace. 
Might be a good idea to either optimize the data fetches and/or add a visual indicator for the initial loading of the data.

wg102:
This is for the two visible tables on jupyter-api. The notebooks and volumes. Might no longer  be an issue, but it also might.

wg102:
Current plan is to add a boolean value that gets set to true the very first time data is received, and by default it will show a loader until it does.
This way, it covers the 'how long until we show a message' question.

wg102:
![image](https://github.com/StatCan/jupyter-apis/assets/14837042/960f9865-5584-4d7a-82de-fb3d55166049)

PR #292, opened 2023-09-26
Feat spinner

wg102:
Closes https://github.com/StatCan/jupyter-apis/issues/257

Need to figure out how to test

### DIFF
--- a/frontend/jupyter/cypress/e2e/main-page.cy.ts
+++ b/frontend/jupyter/cypress/e2e/main-page.cy.ts
@@ -8,7 +8,6 @@ describe('Main table', () => {
     });
     cy.fixture('notebooks').as('notebooksRequest');
     cy.fixture('pvcs').as('pvcsRequest');
-    const notebookInfoLoaded = true;
     cy.visit('/');
   });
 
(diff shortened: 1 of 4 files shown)


wg102:
It looks like this in LOCAL 
image

frontend/jupyter/cypress/e2e/main-page.cy.ts
```diff
@@ -8,7 +8,7 @@ describe('Main table', () => {
     });
     cy.fixture('notebooks').as('notebooksRequest');
     cy.fixture('pvcs').as('pvcsRequest');
-
+    const notebookInfoLoaded = true;
```
mathis-marcotte, line 11: Although this might fix failing tests, it is not how they should be fixed as this variable has nothing to do with the tests or any mock data given to tests, and defining a variable like this is most likely not best practices. That variable should be handled by the component it is used in. The tests are probably failing for another reason that we should investigate.

frontend/jupyter/src/app/pages/index/index-default/index-default.component.html
```diff
@@ -1,76 +1,81 @@
...
-
-  <!--scrollable page content-->
-  <div class="page-padding lib-flex-grow">
-    <lib-table
-      [config]="config"
-      [data]="processedData"
-      [trackByFn]="notebookTrackByFn"
-      (actionsEmitter)="reactToAction($event)"
-    ></lib-table>
+  <div *ngIf="!notebookInfoLoaded">
```
mathis-marcotte, line 2: the spinner should probably appear below the namespace selector, and the namespace select should probably be outside of the new ng-container, as the spinner is just about when the notebooks are loading, not the namespace information (and this was probably the cause of the cypress tests initially failing).

mathis-marcotte (approved):
lgtm

PR #300, opened 2023-10-31
Open new item

wg102:
Closes https://github.com/StatCan/aaw/issues/1782

### DIFF
--- a/frontend/jupyter/src/app/pages/form/form-new/form-data-volumes/form-data-volumes.component.ts
+++ b/frontend/jupyter/src/app/pages/form/form-new/form-data-volumes/form-data-volumes.component.ts
@@ -51,7 +51,7 @@ export class FormDataVolumesComponent implements OnInit {
 
     volGroup.get('mount').setValue(`/home/jovyan/vol-${this.volsArray.length}`);
     volGroup.get('mount').markAsTouched();
-    this.newestItem= volId;
+    this.newestItem = volId;
   }
 
   attachExistingVolume() {
--- a/frontend/jupyter/src/app/pages/form/form-new/form-data-volumes/form-data-volumes.component.ts
+++ b/frontend/jupyter/src/app/pages/form/form-new/form-data-volumes/form-data-volumes.component.ts
@@ -21,6 +21,7 @@ export class FormDataVolumesComponent implements OnInit {
   @Input() readonly: boolean;
   @Input() externalName: string;
   @Input() mountedVolumes: Set<string>;
+  newestItem = 0;
 
   getVolumeTitle = getVolumeTitle;
   getVolumeName = getVolumeName;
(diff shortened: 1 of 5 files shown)


frontend/jupyter/src/app/pages/form/form-new/form-workspace-volume/form-workspace-volume.component.ts
```diff
@@ -52,6 +52,8 @@ export class FormWorkspaceVolumeComponent implements OnInit, OnDestroy {
     this.volGroup.get('mount').setValue('/home/jovyan');
     this.volGroup.enable();
     this.volGroup.get('newPvc.spec.storageClassName').disable();
+    this.panelOpen = true;
+    console.log(this.panelOpen);
```
mathis-marcotte, line 56: console.log should be removed

frontend/jupyter/src/app/pages/form/form-new/form-workspace-volume/form-workspace-volume.component.ts
```diff
@@ -52,6 +52,8 @@ export class FormWorkspaceVolumeComponent implements OnInit, OnDestroy {
     this.volGroup.get('mount').setValue('/home/jovyan');
     this.volGroup.enable();
     this.volGroup.get('newPvc.spec.storageClassName').disable();
+    this.panelOpen = true;
```
mathis-marcotte, line 55: When testing, this one doesn't seem to work for me for workspace volume, either adding a new volume or adding an existing volume.

frontend/jupyter/src/app/pages/form/form-new/form-data-volumes/form-data-volumes.component.ts
```diff
@@ -50,6 +51,7 @@ export class FormDataVolumesComponent implements OnInit {
 
     volGroup.get('mount').setValue(`/home/jovyan/vol-${this.volsArray.length}`);
     volGroup.get('mount').markAsTouched();
+    this.newestItem = volId;
```
mathis-marcotte, line 54: When testing this locally, I get an error in the console
image

To replicate, just click the "add existing data volume button" twice to add two new data volumes.
I think it might just be a case where this line needs to be higher in the function. But that might not be the case.

mathis-marcotte (changes requested):

frontend/jupyter/src/app/pages/form/form-new/form-data-volumes/form-data-volumes.component.ts
mathis-marcotte: Still present after the recent commits, but now noticing this is also happening with main when running it locally. Doesn't seem to happen on dev though. So might be an issue that only firefox returns or something. Reguardless, not an issue with this PR

mathis-marcotte (approved):
lgtm

PR #303, opened 2023-11-09
Block protected b label from configs

wg102:
Part 2 of https://github.com/StatCan/aaw/issues/1871


Jose-Matsuda:
## Test Cases

- [ ] Verify that with just the prob poddefault nothing shows up.
- [ ] Verify that with a prob poddefault and an additional one that the menu shows up, but does not have the prob poddefault. 
- [ ] Verify that we can create a protected b workload and a unprotected workload for both scenarios above

mathis-marcotte (changes requested):
code looks good. But when testing, when going to the form, the configurations dropdown is visible until the first call to getPodDefaults is completed. Then it will show/hide it based on if there are values or not. That is odd behavior since the default podDefaults list is empty, so you end up in a situation where you have no values but the dropdown is still displayed. 
This situation lasts for possibly 1-2 seconds so I am probably being a bit nitpicky, but still, all it takes is a slow network and that becomes more obvious.

All that is needed to do is probably just switch the "isVisible" boolean to default false instead of default true and that should fix this

mathis-marcotte (approved):
lgtm

Issue #307, opened 2024-01-09
Update legacy angular material components


mathis-marcotte:
related to https://github.com/StatCan/jupyter-apis/issues/306

While updating angular versions, I noticed that the angular material components got a major update with v15. https://github.com/StatCan/jupyter-apis/issues/306#issuecomment-1881382688

The automatic update tool updated our imports to use the legacy version for all those components. But, those components are deprecated and until work until angular/material v16. Since we want to be fully on v17, we should update those components.

[Here is the migration help page](https://rc.material.angular.io/guide/mdc-migration#2-run-the-migration-tool)

mathis-marcotte:
work tracked and completed in https://jirab.statcan.ca/browse/BTIS-84

PR #308, opened 2024-01-12
Upgrade angular version v12 to v17

mathis-marcotte:
for https://github.com/StatCan/jupyter-apis/issues/307 (and jira https://jirab.statcan.ca/browse/BTIS-79 which includes the subtasks)

FYI Node version 18 is needed instead of 16 like it used to be, since we are more modern now.


frontend/common/kubeflow-common-lib/projects/kubeflow/src/lib/form/rok-url-input/rok-url-input.component.scss
```diff
@@ -19,7 +19,7 @@
     background: none;
     cursor: default;
   }
-
+  /* TODO(mdc-migration): The following rule targets internal classes of form-field that may no longer apply for the MDC version. */
```
Souheil-Yazji, line 22: Is there an issue for this?

frontend/common/kubeflow-common-lib/projects/kubeflow/src/lib/form/name-namespace-inputs/name-input/name-input.component.spec.ts
```diff
@@ -1,7 +1,7 @@
 import { ComponentFixture, TestBed, waitForAsync } from '@angular/core/testing';
 
 import { NameInputComponent } from './name-input.component';
-import { FormControl } from '@angular/forms';
+import { UntypedFormControl } from '@angular/forms';
```
Souheil-Yazji, line 4: did the typed form produce issues? otherwise we can stick to typed forms https://angular.io/guide/typed-forms#overview-of-typed-forms

frontend/common/kubeflow-common-lib/projects/kubeflow/src/styles.scss
```diff
@@ -1,8 +1,17 @@
-@use '~@angular/material' as mat;
+@use '@angular/material' as mat;
 @import './kubeflow.css';
 @import './lib/variables.scss';
 
 // always include only once per project
+// TODO(v15): As of v15 mat.legacy-core no longer includes default typography styles.
```
Souheil-Yazji, line 6: Is there an issue for this?

frontend/common/kubeflow-common-lib/projects/kubeflow/tsconfig.lib.json
```diff
@@ -3,7 +3,6 @@
   "compilerOptions": {
     "outDir": "../../out-tsc/lib",
     "declarationMap": true,
-    "target": "es2015",
```
Souheil-Yazji, line 6: why no more target version? maybe `es2022` like below config?

frontend/jupyter/cypress/e2e/form-page.cy.ts
```diff
@@ -129,7 +129,7 @@ describe('New notebook form', () => {
       // not smaller than minimum cpu
       cy.get('[data-cy-form-input="cpu"]').clear();
       cy.get('[data-cy-form-input="cpu"]').find('input').type('5.0');
-      cy.get('[data-cy-form-input="cpuLimit"]').find('input').type('4.0');
+      cy.get('[data-cy-form-input="cpuLimit"]').find('input').type('4.0', {force: true});
```
Souheil-Yazji, line 132: whatcha doin?

frontend/common/kubeflow-common-lib/projects/kubeflow/src/lib/form/rok-url-input/rok-url-input.component.scss
mathis-marcotte: Its from the automated script to migrate the legacy angular components https://jirab.statcan.ca/browse/BTIS-84 . But its in an "rok" component, which is not a part of the repo that we actually use. So there's no real need to update this. (Honestly the entire rok stuff should just get removed)

frontend/common/kubeflow-common-lib/projects/kubeflow/src/lib/form/name-namespace-inputs/name-input/name-input.component.spec.ts
mathis-marcotte: This was done automatically when upgrading to angular 14 https://angular.io/guide/typed-forms#automated-untyped-forms-migration. But i agree that we should stick to typed forms to be more robust. I can create an issue to do the transition to typed forms

frontend/common/kubeflow-common-lib/projects/kubeflow/src/styles.scss
mathis-marcotte: Ok good catch for the todo. But It's actually probably most likely irrelevant now. It was introduced during the angular v15 update, which was the update that brought a lot of changes to the components and was the update that flagged a lot of our components as "legacy" which required me to run the migration script to update all the legacy components.
Pre migration script:
image

Post migration script: 
image


If the todo is irelevent now though, it should be removed so I will go and do that.

frontend/common/kubeflow-common-lib/projects/kubeflow/tsconfig.lib.json
mathis-marcotte: This actually got updated to es2020 when i updated to angular 14. But then got deleted when i updated to v15. Not sure why 100%, Ill look into the cause for this because you are right that this might be suspicious. 

Oh actually now that I take a look at it, this file extends "../../tsconfig.json", which has the "target" property already(with value es2022)

frontend/jupyter/cypress/e2e/form-page.cy.ts
mathis-marcotte: Ok so this one is a bit weird, following the component updates, cypress would just fail at some of these `type` commands. From what I can tell, this comes from how the new components are getting generated with the "required/mandatory" validator, as we now have the mandatory symbol "*" with our inputs. I think that although visually it doesn't appear this way, I think that this mandatory aspect is being applied OVER the actual input component. So when cypress tries to find this input, it can find it, but it thinks it's being covered by something, so it doesn't want to type what I tell it to type. So that is why I need to force it to type in the input

frontend/jupyter/src/app/pages/form/form-new/volume/existing/pvc/pvc.component.scss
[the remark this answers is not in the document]
mathis-marcotte: it was looking a bit wonky, inputs were too close to each other, plus now we actually display the size of the existing volumes since i fixed that code,

PR #312, opened 2024-01-22
Add /default api

wg102:
Part of Jira https://jirab.statcan.ca/browse/BTIS-77 to show the default notebook

### DIFF
--- a/notebooks.go
+++ b/notebooks.go
@@ -315,54 +315,7 @@ func (s *server) GetNotebooks(w http.ResponseWriter, r *http.Request) {
 	}
 
 	for _, notebook := range notebooks {
-		// Load events
-		allevents, err := s.listers.events.Events(notebook.Namespace).List(labels.Everything())
-		if err != nil {
-			log.Printf("failed to load events for %s/%s: %v", notebook.Namespace, notebook.Name, err)
-		}
-
-		// Filter past events
-		events := make([]*corev1.Event, 0)
-		for _, event := range allevents {
-			if event.InvolvedObject.Kind != "Notebook" || event.InvolvedObject.Name != notebook.Name || event.CreationTimestamp.Before(&notebook.CreationTimestamp) {
-				continue
-			}
-
-			events = append(events, event)
-		}
-		sort.Sort(eventsByTimestamp(events))
-
-		imageparts := strings.SplitAfter(notebook.Spec.Template.Spec.Containers[0].Image, "/")
-
-		// Process current status + reason
-		status := processStatus(notebook, events)
-
-		volumes := []string{}
-		for _, vol := range notebook.Spec.Template.Spec.Volumes {
-			volumes = append(volumes, vol.Name)
-		}
-
-		cpulimit := resource.Zero.AsDec()
-		if req, ok := notebook.Spec.Template.Spec.Containers[0].Resources.Requests[corev1.ResourceCPU]; ok {
-			cpulimit = req.AsDec()
-		}
-
-		resp.Notebooks = append(resp.Notebooks, notebookresponse{
-			Age:          notebook.CreationTimestamp.Time,
-			Name:         notebook.Name,
-			Namespace:    notebook.Namespace,
-			Image:        notebook.Spec.Template.Spec.Containers[0].Image,
-			LastActivity: notebook.Annotations[LastActivityAnnotation],
-			ServerType:   notebook.Annotations[ServerTypeAnnotation],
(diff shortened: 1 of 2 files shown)


notebooks.go
```diff
@@ -363,6 +368,88 @@ func (s *server) GetNotebooks(w http.ResponseWriter, r *http.Request) {
...
+		APIResponseBase: APIResponseBase{
+			Success: false,
+			Status:  http.StatusOK,
+		},
+	}
+
+	for _, notebook := range notebooks {
+
+		if val, ok := notebook.Labels["notebook.statcan.gc.ca/default-notebook=true"]; ok {
+			if val == "true" {
```
mathis-marcotte, line 393: What is this for?

notebooks.go
```diff
@@ -363,6 +368,88 @@ func (s *server) GetNotebooks(w http.ResponseWriter, r *http.Request) {
...
+				Volumes:      volumes,
+				Labels:       notebook.Labels,
+				Metadata:     notebook.ObjectMeta,
+			}
+			s.respond(w, r, resp)
+			return
+		}
+	}
+
+	s.respond(w, r, resp)
```
mathis-marcotte, line 450: you have two responds for one function which can cause confusions and can easily be avoided. For something like this, I would suggest doing something more like:
```
for each notebook{
    if default notebook{
        default notebook found
        break;
   }
}
process notebook if found;
return response;
```

notebooks.go
```diff
@@ -139,6 +139,11 @@ type notebooksresponse struct {
 	Notebooks []notebookresponse `json:"notebooks"`
 }
 
+type notebookapiresponse struct {
```
mathis-marcotte, line 142: Is there a reason you need that specific type of notebook response? Can't you just use the "getnotebookresponse" that is just underneath, and did something like the getNotebook function that we already have https://github.com/StatCan/jupyter-apis/blob/63ceeaca2ff2f85faf2a962acb90fd0a7844c075/notebooks.go#L1065?

mathis-marcotte (changes requested):

notebooks.go
wg102: It's me forgetting to remove the second if while i was testing

notebooks.go
wg102: cause i want the normal data and the api response. and the other one has a kybeflow format, this one has our final

notebooks.go
wg102: good idea

notebooks.go
mathis-marcotte: makes sense

notebooks.go
```diff
@@ -363,6 +368,85 @@ func (s *server) GetNotebooks(w http.ResponseWriter, r *http.Request) {
...
+			Status:  http.StatusOK,
+		},
+	}
+
+	for _, notebook := range notebooks {
+
+		if val, ok := notebook.Labels["notebook.statcan.gc.ca/default-notebook"]; ok {
+			if val == "true" {
+				// Load events
+				allevents, err := s.listers.events.Events(notebook.Namespace).List(labels.Everything())
```
mathis-marcotte, line 395: I'm not a fan of the large code duplication that is happening between this new function and the GetNotebooks function https://github.com/StatCan/jupyter-apis/blob/ede3c5adb70a427437676493dfa4bd7deccdc7f1/notebooks.go#L295.

What you should do is transfer all that code that is in the for loops for both those functions to put in in its own function "processNotebook(notebook) that returns a notebookresponse object. Then, inside those for loops, you would just need to call that function, get the notebookresponse, and put that in the resp.Notebook.

I'm just thinking that if we ever need to change how we process notebooks, it will be a lot easier to do if the code is all in one place.

mathis-marcotte (changes requested):

notebooks.go
```diff
@@ -310,57 +315,91 @@ func (s *server) GetNotebooks(w http.ResponseWriter, r *http.Request) {
...
+				resp.Notebook = getNotebookData(notebook, s)
+				resp.APIResponseBase.Success = true
+				break
+			}
 		}
+	}
+	s.respond(w, r, resp)
+}
+
+func getNotebookData(notebook *kubeflowv1.Notebook, s *server) notebookresponse {
```
mathis-marcotte, line 355: can you declare the `s *server` like all the other functions in this file? thanks!

mathis-marcotte (changes requested):

mathis-marcotte (approved):
lgtm

PR #313, opened 2024-01-31
Add the 2.0 branch

wg102:


.github/workflows/build.yml
```diff
@@ -4,6 +4,7 @@ on:
   pull_request:
     branches:
       - main
```
mathis-marcotte, line 6: shoulb remove main probably if we dont need that reference, same goes for the other file

mathis-marcotte (changes requested):

mathis-marcotte (approved):
lgtm

PR #323, opened 2024-05-03
added fix to enumerate duplicate names in default notebook creation

mathis-marcotte:
fixes https://jirab.statcan.ca/browse/BTIS-270


notebooks.go
Souheil-Yazji, line 1: I think this is looking for whether the pvc exists, then adding a number to the name according to the count, then using that name to create the storage. It's also doing something similar for notebook names. 

Can I get you to add comments to your code? Probably specifying why we use enumarateNames, in-line or function level is up to you :)

Souheil-Yazji (changes requested):
LGTM!

notebooks.go
mathis-marcotte: The thought had crossed my mind at some point, and then I forgot about it. But good call, will do!

Souheil-Yazji (approved):

PR #325, opened 2024-06-26
Replaced Dockle with hadolint

mathis-marcotte:
for https://jirab.statcan.ca/browse/BTIS-390


wg102 (commented):
Any reason why we now hadolint before local build?

wg102 (approved):

mathis-marcotte:
Following how it's done in kubeflow-containers. Makes a lot more sense to lint the dockerfile before building it if there are serious issues in there.

PR #331, opened 2024-11-26
label buttons

AbdallaAbdelhadi:


mathis-marcotte:
The build is failing here, trivy is reporting a vulnerability. I think that is something we can fix. I think we just need to update our golang version to 1.21 from 1.20. This could be done in a seperate PR.

PR #332, opened 2024-12-11
chore(zone-dev): update minimum req

Jose-Matsuda:
https://jirab.statcan.ca/browse/BTIS-603

We can change the default but we need to also make this change to pass validation


mathis-marcotte (approved):
lgtm

PR #337, opened 2025-02-18
refactor: changing settings in jupyter ui components to reflect scali…

jacek-dudek:
…ng down of CPU and memory resources for notebooks.


mathis-marcotte (approved):
LGTM. Just make sure to update the manifest spawner ui config when you update the manifest, if you want to edit the default values. https://github.com/StatCan/aaw-kubeflow-manifests/blob/aaw-dev-cc-00/kustomize/apps/jupyter-web-app/base/configs/spawner_ui_config.yaml#L68

PR #339, opened 2025-02-25
Kf1.9 init common update

mathis-marcotte:
for https://jirab.statcan.ca/browse/BTIS-663


mathis-marcotte:
I am aware that the JWA Frontend Tests, as well as the publish workflows are failing. That is to be expected.
This PR only updates what is found under the `./frontend/common/kubeflow-common-lib` folder. 
The reason those workflows are failing, and the reason I can't fully run the app yet, is that code also needs to be updated under `./frontend/jupyter`, to follow with the updates that were made to the common-lib. 

But as we can see, the workflows for "Common Frontend Tests" still all successfully pass, those which actually touches the code I updated..

wg102 (approved):
LGTM

PR #344, opened 2025-02-28
feat: Replace thunder testing

briansu2004:
JupyterAPI - Replace Thunder Testing
https://jirab.statcan.ca/browse/BTIS-660


rest-tests/restclient.http
```diff
@@ -0,0 +1,94 @@
+@KF_USER_ID=briangengda.su@statcan.gc.ca
```
mathis-marcotte, line 1: Two things:

1.  You might want to remove the values you set for these variables. Ideally we would have a testing account for this, but we don't. And then that way, whenever someone wants to use these tests, they can add values to these variables on their local clone. 
I mainly just want to avoid you having your email commited in the repo since this is technically a public repo, just to be safe.

2. Is there any way to add an http header by default to these calls? I am getting a lot of 404s
```
{"success":false,"status":404,"log":"no user found","user":""}
```
And that is because the `kubeflow-userid` http header is missing from these tests. So you would need to set this http header of `kubeflow-userid` with the value from the @KF_USER_ID variable to solve this "no user found" error.

I didn't really look at the REST client docs yet to figure out how to add http headers. But I just quickly booted up postman and to confirm my theory, here is an example of it returning a 200 status, so that you can see it working
image

mathis-marcotte (changes requested):
Just a couple things to try and make this a little bit cleaner.

But first impressions, I like how easy it is to look through the .http file.

rest-tests/restclient.http
briansu2004: Updated PR

Souheil-Yazji (approved):

PR #345, opened 2025-03-04
fix: sas default tag

EveningStarlight:
The SAS default tag is `latest`, when it should be `v1`. `latest` is shared with the Zone, and `v1` is for AAW exclusively. 

This means that the image pulled isn't stable or reliably using the AAW image.


mathis-marcotte (approved):
lgtm

PR #348, opened 2025-03-10
update(KF1.9): Updated frontend volumes related component

mathis-marcotte:
for https://jirab.statcan.ca/browse/BTIS-836

Updates the volumes components we import from the upstream volumes app.

Also fixes the bugs with the kf1.9 update for jupyter-apis, just enough to get the frontend running. But no further testing done as that is it's own dedicated task.
(So the build workflow should be successful)


frontend/jupyter/src/app/pages/index/index-default/index-default.component.ts
```diff
@@ -406,20 +409,79 @@ export class IndexDefaultComponent implements OnInit, OnDestroy {
...
-            SnackType.Warning,
-            3000,
-          );
+          const config: SnackBarConfig = {
+            data: {
+              msg: 'PVC is unavailable now.',
+              snackType: SnackType.Warning,
+            },
+          };
+          this.snackBar.open(config);
```
wg102, line 484: How come this isn't localized while line 500 is? 
I kinda assume localization is in another ticket but if they are both snackbars I would think they behave the same way

frontend/jupyter/src/app/services/actions.service.ts
```diff
@@ -143,7 +143,13 @@ export class ActionsService {
...
             const object = `${namespace}/${name}`;
             const message = $localize`Delete request was sent.`;
-            this.snackBar.open(`${object}: ${message}`, SnackType.Info, 3000);
+            const config: SnackBarConfig = {
+              data: {
+                msg: `${object}: ${message}`,
+                snackType: SnackType.Info,
+              },
+            };
+            this.snackBar.open(config);
```
wg102, line 152: Probably not a part of the review, but have you noticed the amount of snackbar/snackbarconfigs and yet we never set them quite the same way? 
Assuming it's an upstream thing but wouldn't it be cleaner to make a method or service and have them all call it so it's the same code?

frontend/jupyter/src/app/types/volume/interfaces.ts
```diff
@@ -35,6 +39,8 @@ export interface Volume {
 export interface PVCProcessedObject extends PVCResponseObject {
   deleteAction?: string;
   editAction?: string;
+  closePVCViewerAction?: string;
+  openPVCViewerAction?: string;
```
wg102, line 43: Once again upstream, but why need two? Wouldnt the action be either open or closed? Unless they can be both at once?

frontend/jupyter/src/app/pages/index/index-default/index-default.component.ts
mathis-marcotte: I dont know. It was like that before. This just changes how the snackbars are generated.

I already made note in the translation ticket to review every snackback message.

frontend/jupyter/src/app/services/actions.service.ts
mathis-marcotte: The snackbar config is the way that upstream standardized the snackbar changes. 

Not sure what you mean by us setting them differently, I didn't see any major difference between us and upstream in terms of snackbars, outside of the localized messages.

frontend/jupyter/src/app/types/volume/interfaces.ts
mathis-marcotte: string might not be a boolean. I haven't seen the pvc viewer in action since we dont have that in our cluster yet.
There might be multiple states for these actions.

wg102:
LGTM

wg102 (approved):

PR #350, opened 2025-03-20
feat(jupyter): Remove limits for UI

wg102:
https://jirab.statcan.ca/browse/BTIS-841
image


Jose-Matsuda (approved):
tested on aaw dev (zone dev was unsure about the status of testability) and after sending a few create notebook requests just confirmed that limit and requests are tied to each other

PR #351, opened 2025-03-24
fix: removed vulnerabilities for common-lib and jupyter

mathis-marcotte:
for https://jirab.statcan.ca/browse/BTIS-672

### DIFF
--- a/frontend/common/kubeflow-common-lib/package.json
+++ b/frontend/common/kubeflow-common-lib/package.json
@@ -81,6 +81,9 @@
     "http-proxy-middleware": "2.0.7",
     "webpack-dev-middleware": "6.1.2",
     "undici": "6.21.1",
-    "webpack": "5.94.0"
+    "webpack": "5.94.0",
+    "@babel/runtime": "7.26.10",
+    "esbuild": "0.25.1",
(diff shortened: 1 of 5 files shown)


frontend/common/kubeflow-common-lib/package.json
```diff
@@ -81,6 +81,9 @@
     "http-proxy-middleware": "2.0.7",
     "webpack-dev-middleware": "6.1.2",
     "undici": "6.21.1",
-    "webpack": "5.94.0"
+    "webpack": "5.94.0",
+    "@babel/runtime": "7.26.10",
+    "esbuild": "0.25.1",
+    "vite": "5.4"
```
wg102, line 87: Why need to add those 3? 
I see they are in the jupyter package.json and overwritten but are they needed here too?

wg102 (approved):
One comment but LGTM

frontend/common/kubeflow-common-lib/package.json
mathis-marcotte: fixes the remaining vulns in the common lib. 
Since they are the same as the ones coming from jupyter, I figured might as well fix them here as well.

PR #352, opened 2025-03-25
Kf1.9(jupyter): fix style

wg102:
See https://jirab.statcan.ca/browse/BTIS-882

### DIFF
--- a/frontend/common/kubeflow-common-lib/projects/kubeflow/src/styles/styles.scss
+++ b/frontend/common/kubeflow-common-lib/projects/kubeflow/src/styles/styles.scss
@@ -5,6 +5,7 @@
 // Define the primary palette for the color that appears most
 // frequently throughout your application.
 $kubeflow-primary-palette: (
+  100: $kubeflow-blue-hue-1,
   400: $kubeflow-blue-hue-1,
   500: $kubeflow-blue-hue-2,
   600: $kubeflow-blue-hue-3,
@@ -55,11 +56,10 @@ $kubeflow-typography-config: mat.define-typography-config(
 // theming systems such as "color" or "typography".
 $kubeflow-theme: mat.define-light-theme(
   (
-    //TODO: AAW figure out why this breaks
-    // color: (
-    //   primary: mat.define-palette($kubeflow-primary-palette, 700),
-    //   accent: mat.define-palette(mat.$pink-palette, A200, A100, A400),
-    // ),
+    color: (
+      primary: mat.define-palette($kubeflow-primary-palette, 700),
+      accent: mat.define-palette(mat.$pink-palette, A200, A100, A400),
+    ),
     typography: $kubeflow-typography-config
   )
 );
@@ -86,9 +86,7 @@ $kubeflow-theme: mat.define-light-theme(
 }
 
 // Use the material icons offline
-// AAW: Leaving this blank was the way to make this work
-// after angular version update without returning errors
-$material-icons-font-path: '';
+$material-icons-font-path: '~material-icons/iconfont/';
 @import 'node_modules/material-icons/iconfont/material-icons.scss';
 
 // Custom css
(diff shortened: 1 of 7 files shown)


frontend/common/kubeflow-common-lib/projects/kubeflow/src/styles/styles.scss
```diff
@@ -86,9 +86,7 @@ $kubeflow-theme: mat.define-light-theme(
 }
 
 // Use the material icons offline
-// AAW: Leaving this blank was the way to make this work
-// after angular version update without returning errors
-$material-icons-font-path: '';
+$material-icons-font-path: '~material-icons/iconfont/';
```
mathis-marcotte, line 89: Wow I swear this used to break things somewhere. But this doesnt seem to break anything so that is good.

frontend/common/kubeflow-common-lib/projects/kubeflow/src/styles/styles.scss
```diff
@@ -5,6 +5,7 @@
 // Define the primary palette for the color that appears most
 // frequently throughout your application.
 $kubeflow-primary-palette: (
+  100: $kubeflow-blue-hue-1,
```
mathis-marcotte, line 8: I assume this is what was missing to make that TODO that I had put finally work.

frontend/jupyter/src/app/pages/index/index-default/config.ts
```diff
@@ -152,7 +152,7 @@ export const defaultConfig: TableConfig = {
         new ActionIconValue({
           name: 'delete',
           tooltipReady: $localize`Delete this notebook server`,
-          color: '',
+          color: 'warn',
```
mathis-marcotte, line 155: Seeing it live, I don't know if this looks good.
Should we instead try to keep our icon colors uniform? And instead of making this one red, should we instead make the volumes delete button black?

frontend/jupyter/src/styles.scss
```diff
@@ -4,7 +4,7 @@
 
 // Give some space after each input
 mat-form-field {
-  margin-bottom: 0.25rem;
+  margin-bottom: 0rem;
```
mathis-marcotte, line 7: Is there a reason for this change? upstream has this styling like it was.

mathis-marcotte (changes requested):
Just some small questions/comments

frontend/common/kubeflow-common-lib/projects/kubeflow/src/styles/styles.scss
wg102: Yes it was just not there.

frontend/jupyter/src/styles.scss
wg102: That is why we had a double line. I know upstream has it like that, my best guess it that the material ui has changed just a tiny bit.

frontend/jupyter/src/app/pages/index/index-default/config.ts
wg102: I like the standard part of it. It's not like it's something hard to change down the line either. So this one is more of a choice.

frontend/jupyter/src/styles.scss
mathis-marcotte: We've always had that double line, it's currently live on prod. I don't think this is the right idea to fix to fix something like that since this impacts the entire form page, even if it is very minor white space getting removed.
The double line issue seems like a separate issue based on the components we are using. Not on the generic kubeflow styling. 
Like on that topic, we also have a different background color than upstream for those input boxes in the dashboard page.
I think we should keep the same styling as upstream here, and make a separate dedicated ticket for fixing those filter input boxes, since this problem is not related to kf1.9.

frontend/jupyter/src/app/pages/index/index-default/config.ts
mathis-marcotte: Can we make the choice now then to update the volume's delete buttons to black? Instead of keeping this change for the notebook's delete icon to red.

wg102: Decision by TL is set all the delete to red for "dangerous".

frontend/jupyter/src/styles.scss
wg102: Reverted the change. Creating a follow-up ticket for this

mathis-marcotte (approved):
lgtm

PR #353, opened 2025-03-25
Kf1.9(jupyter): fix i18n

wg102:
https://jirab.statcan.ca/browse/BTIS-674

### DIFF
--- a/frontend/common/kubeflow-common-lib/projects/kubeflow/src/lib/form/validators.ts
+++ b/frontend/common/kubeflow-common-lib/projects/kubeflow/src/lib/form/validators.ts
@@ -20,8 +20,8 @@ export const dns1123Validator: IValidator = {
   regex: '^' + dns1123LabelFmt + '(\\.' + dns1123LabelFmt + ')*' + '$',
   help:
     // prettier-ignore
-    'Name must consist of lowercase alphanumeric characters or \'-\', and"' +
-    ' must start and end with an alphanumeric character',
+    $localize `Name must consist of lowercase alphanumeric characters or '-',
+      and must start and end with an alphanumeric character`,
 };
 
 // TODO(kimwnasptd): We only use this validator, do we need the others?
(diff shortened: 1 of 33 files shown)


frontend/common/kubeflow-common-lib/projects/kubeflow/src/lib/form/validators.ts
```diff
@@ -30,12 +30,11 @@ export const dns1035Validator: IValidator = {
   help: $localize`Name must consist of lowercase alphanumeric characters or '-',
     start with an alphabetic character, and end with an alphanumeric character.`,
 };
-
+// aaw verify if i18n works
```
Souheil-Yazji, line 33: comment isn't clear, it seems all we did here is add the i18n. You want us to test if it works?

frontend/jupyter/package.json
```diff
@@ -41,7 +41,7 @@
     "@fortawesome/fontawesome-svg-core": "^1.2.35",
     "@fortawesome/free-brands-svg-icons": "^5.15.3",
     "@fortawesome/free-solid-svg-icons": "^5.15.3",
-    "@kubernetes/client-node": "^0.16.3",
+    "@kubernetes/client-node": "^1.0.0",
```
Souheil-Yazji, line 44: Looks good.
https://github.com/kubernetes-client/javascript?tab=readme-ov-file#compatibility

frontend/jupyter/src/app/pages/index/index-default/server-type/server-type.component.html
Souheil-Yazji, line 1: makes sense for aaw, not the zone

Souheil-Yazji (approved):
See comments, LGTM

frontend/jupyter/src/app/pages/index/index-default/server-type/server-type.component.html
wg102: Since the first one is using it, I figured there was no hurt in finishing the rest. This being said this specific thing doesn't show correctly when local build. Creating a ticket to fix it.

frontend/common/kubeflow-common-lib/projects/kubeflow/src/lib/form/validators.ts
wg102: Forgot to remove... REmoving

PR #359, opened 2025-04-24
fix: Fix various kf1.9 issues from testing 

mathis-marcotte:
for https://jirab.statcan.ca/browse/BTIS-852


frontend/jupyter/src/app/pages/index/index-default/index-default.component.ts
Souheil-Yazji, line 1: I don't like how the error event no longer exists here.

status.go
Souheil-Yazji, line 1: this is a healthy refactor, minus the comment left above.

Souheil-Yazji (approved):

PR #364, opened 2025-07-17
fix(notebook form): Fixed error popup about namespaces in new notebook form

mathis-marcotte:
for https://jirab.statcan.ca/browse/ZONE-113

The PR is a bit big for what I'm really fixing.

The main root of the issue was the call that I removed from `frontend/jupyter/src/app/pages/form/form-new/form-new.component.ts`

But I didn't like the idea of just removing the problem there, while letting a bunch of dead and related code lying around. So I took this time to remove the useless function from the backend and frontend completely instead of just removing it from where it was called.

### DIFF
--- a/frontend/jupyter/i18n/fr/messages.fr.xlf
+++ b/frontend/jupyter/i18n/fr/messages.fr.xlf
@@ -580,7 +580,7 @@
         </context-group>
         <context-group purpose="location">
           <context context-type="sourcefile">src/app/services/backend.service.ts</context>
-          <context context-type="linenumber">325</context>
+          <context context-type="linenumber">317</context>
         </context-group>
       </trans-unit>
       <trans-unit id="3862207452684082108" datatype="html">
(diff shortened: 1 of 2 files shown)


JustinZhang17:
I was wondering if the zone currently uses namespace level restrictions at all? Also what is image group three?

This change is an entire policy change and our stakeholders/end users would need to know that namespace restrictions are lifted.

frontend/jupyter/src/app/pages/form/form-new/form-cpu-ram/form-cpu-ram.component.ts
```diff
@@ -103,21 +101,6 @@ export class FormCpuRamComponent implements OnInit, OnChanges {
...
-    ) {
-      this.MAX_FOR_GPU = new Map([
-        [0, { cpu: 4, memory: 24, cpuLimit: 8, memoryLimit: 32 }],
-      ]);
-    } else {
-      this.MAX_FOR_GPU = new Map([
-        [0, { cpu: 14, memory: 48, cpuLimit: 14, memoryLimit: 48 }],
-        [1, { cpu: 4, memory: 96, cpuLimit: 4, memoryLimit: 96 }],
-      ]);
-    }
```
JustinZhang17, line 120: I thought this was a complete change in functionality, but all this code removed is just a relic of aaw

PR #368, opened 2025-12-23
Update the name of group-three to sas

wg102:
https://jirab.statcan.ca/browse/BTIS-561


frontend/jupyter/src/app/pages/index/index-default/server-type/server-type.component.html
```diff
@@ -20,8 +20,8 @@
   i18n-aria-label
 ></mat-icon>
 <mat-icon
-  *ngIf="notebookServerType === 'group-three'"
-  svgIcon="group-three-icon"
+  *ngIf="notebookServerType === 'sas' || notebookServerType === 'group-three'"
```
mathis-marcotte: We already talked about this, but I will just write it on the PR so we don't lose it.

Instead of doing an "or serverType = group-three" logic in jupyter-apis, let's instead make a script to update all the old notebooks to use the new name for their serverType.

Will make a new jira ticket/sub-task to work on that. and We can run that script when we update this code. (worse case, old sas notebooks won't have the SAS icon displayed in the UI for a few minutes)

mathis-marcotte (changes requested):

mathis-marcotte (approved):
lgtm

PR #376, opened 2026-02-03
feat(all): delay culling

wg102:
Adding the dialog to add a time for culling. 

The functionality works and i18n is done.
Validation and test done

I18n will need to be updated when ready to merge since other PRS have had a deep impact


bryanpaget (approved):
lgtm (for alpha) :-)

rest-tests/restclient.http
```diff
@@ -1,6 +1,6 @@
 ### Replace these placeholders with actual values
-@KF_USER_ID=REPLACE_WITH_YOUR_USER_ID
-@KF_NAMESPACE=REPLACE_WITH_YOUR_NAMESPACE
+@KF_USER_ID=wendy.gaultier@statcan.gc.ca
```
mathis-marcotte: Remove your name from this and keep the default values when making commits

rest-tests/restclient.http
```diff
@@ -158,3 +158,11 @@ kubeflow-userid: {{KF_USER_ID}}
 {
     "name": "{{KF_PVC_VIEWER_NAME_CREATE}}"
 }
+
+### Postpone Culling
+### api/namespaces/${namespace}/notebooks/${name}/keepalive/
```
mathis-marcotte: you can probably remove this line from the comments to keep this file clean, we don't need to see the call url when we can see it right underneath

notebooks.go
```diff
@@ -46,6 +46,8 @@ const AutoMountLabel string = "data.statcan.gc.ca/inject-blob-volumes"
 
 // LastActivityAnnotation is the annotation name for the last activity value.
 const LastActivityAnnotation = "notebooks.kubeflow.org/last-activity"
+// LastActivityCheckTimeStamp for the delay shutdown
+const LastActivityCheckTimeStamp = "notebooks.kubeflow.org/last-activity"
```
mathis-marcotte: Can you not just reuse the const that is right above this?

notebooks.go
```diff
@@ -1046,12 +1052,75 @@ func formatMemoryToGibibytes(memory resource.Quantity) string {
...
+	})
+}
+
 func (s *server) GetNotebook(w http.ResponseWriter, r *http.Request) {
 	vars := mux.Vars(r)
 	namespace := vars["namespace"]
 	notebook := vars["notebook"]
 
-	log.Printf("getting notebook %q for %q", notebook, namespace)
+	// log.Printf("getting notebook %q for %q", notebook, namespace)
```
mathis-marcotte: Is there a reason you are commenting this out? I think we can keep this logging message.

notebooks.go
```diff
@@ -1046,12 +1052,75 @@ func formatMemoryToGibibytes(memory resource.Quantity) string {
 	return strconv.FormatFloat(gibibytes, 'f', -1, 64)
 }
 
+func (s *server) UpdateNotebookForCulling(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	namespaceName := vars["namespace"]
+	notebookName := vars["notebook"]
+	log.Printf("updating notebook %q for %q with additional time", notebookName, namespaceName)
+	// Todo: validate that the keepalive value is valid.
```
mathis-marcotte: Seems like you are calling the validation on line 1077, so this TODO comment can probably be removed?

notebooks.go
```diff
@@ -1046,12 +1052,75 @@ func formatMemoryToGibibytes(memory resource.Quantity) string {
...
+		return
+	}
+
+	numKeepAliveTime, err := strconv.Atoi(req.TimeHours)
+	if err != nil {
+		fmt.Println("Error while parsing:", err)
+		return
+	}
+	updatedTime := time.Now().Add(time.Duration(numKeepAliveTime) * time.Hour)
+	log.Printf("Updated time notebook %q", updatedTime);
```
mathis-marcotte: maybe it might be a more complete log message to also add the time in there? Like "updated notebook %notebook last activity time to %time"? Open to other suggestions too

notebooks.go
```diff
@@ -1046,12 +1052,75 @@ func formatMemoryToGibibytes(memory resource.Quantity) string {
...
+	if notebook.Annotations == nil {
+		notebook.Annotations = map[string]string{}
+	}
+
+	notebook.Annotations[LastActivityAnnotation] = updatedTime.Format(time.RFC3339)
+	notebook.Annotations[LastActivityCheckTimeStamp] = updatedTime.Format(time.RFC3339)
```
mathis-marcotte, line 1102: Based on your new const values, these lines both update the `notebooks.kubeflow.org/last-activity` annotation. 

I'm not too sure of what the correct logic is here, but : 

- if one of the consts values is wrong, update it and then it makes sense to have both of these
- if you only need one const, then only patch the value once here

notebooks.go
```diff
@@ -1046,12 +1052,75 @@ func formatMemoryToGibibytes(memory resource.Quantity) string {
...
+	 
+	if true {
```
mathis-marcotte: if true? What is the reason for this here? Remove if not needed

notebooks.go
```diff
@@ -1421,3 +1490,28 @@ func validateNotebookVolume(req volrequest, validsizes map[int64]bool) error {
...
+func validateCullingDelay(request delaycullingrequest) error {
+	var validationErrors []string
+
+	log.Printf("validating delay of %v", request.TimeHours)
+
+	// Required string fields
+	if request.TimeHours == "" {
+		validationErrors = append(validationErrors, "Delay is required")
+	}
+	matched, err := regexp.MatchString(`^([1-9]|[1-6]\d|7[0-2])$`, request.TimeHours)
```
mathis-marcotte: This seems like a really complicated way to validate this value. If we are just expecting a number, maybe we should validate it as such? 

So maybe a better plan would be to do the `strconv.Atoi(req.TimeHours)` step BEFORE the validation.

And then you would pass that value to your validate `func validateCullingDelay(timehours int) error { ... }`

And then instead of using a RegEx for validating an int, you could just do like `if timeHours < 0 || timeHours > 72`. 

How does that sound?

frontend/jupyter/README.md
```diff
@@ -20,6 +20,8 @@ Run `ng test` to execute the unit tests via [Karma](https://karma-runner.github.
 
 ## Running integration tests
 
+Make sure the tests are being run in English. and the namespace is the 'kubeflow-userid'.
```
mathis-marcotte, line 23: (not a change request) Thank you for adding this!

frontend/jupyter/src/app/pages/notebook-page/notebook-page.component.spec.ts
```diff
@@ -33,6 +33,7 @@ const ActionsServiceStub: Partial<ActionsService> = {
   deleteNotebook: () => of(),
   startNotebook: () => of(),
   stopNotebook: () => of(),
+  updateKeepAlive: () => of(),
```
mathis-marcotte: Is this needed here? You aren't editing the notebook-page.component.

frontend/common/kubeflow-common-lib/projects/kubeflow/src/lib/delay-dialog/delay-dialog.component.spec.ts
mathis-marcotte, line 1: Shouldn't you have at least something here for unit tests? 

See the confirm-dialog for examples

frontend/common/kubeflow-common-lib/projects/kubeflow/src/lib/delay-dialog/delay-dialog.module.ts
```diff
@@ -0,0 +1,23 @@
...
+    FormsModule,
+    MatButtonModule,
+    MatDialogModule,
+    MatFormFieldModule,
+    MatInputModule,
+    MatDialogClose,
+    ReactiveFormsModule,
+  ],
+})
+export class DelayDialogModule {}
```
mathis-marcotte, line 25: (Not related to this file specifically, just didn't know where else to comment this)

When I look at the `confirm-dialog` that we already have, I see that it has files for `confirm-dialog.service`, which you don't have it seems. 

Do you know what the difference is or what the use is for that service? Do you think both our dialog types should be designed similarily?

frontend/common/kubeflow-common-lib/projects/kubeflow/src/lib/delay-dialog/delay-dialog.component.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  ngOnInit() {
+    const fb = new FormBuilder();
+
+    this.formDelayCtrl = fb.group({
+      hourControl: [
+        '1',
+        [
+          Validators.required,
+          Validators.min(0),
```
mathis-marcotte: Unless this is an exclusive value, wouldn't it make more sense to have the min value be 1? Delaying the culling by 0 hours won't really do much, you know?

frontend/common/kubeflow-common-lib/projects/kubeflow/src/lib/delay-dialog/delay-dialog.component.ts
```diff
@@ -0,0 +1,69 @@
...
+          Validators.max(72),
+          Validators.pattern('^([1-9]|[1-6][0-9]|7[0-2])$'),
```
mathis-marcotte: This regex seems complicated to follow. 
Since you already have the min and max validator, do you think it's necessary? Or if it is, would it be simpler to just have a pattern to make sure it's just a number, and then let the min/max validators check if it's between 1 and 72? So a regex like `^[0-9]*$` ?

frontend/common/kubeflow-common-lib/projects/kubeflow/src/lib/delay-dialog/delay-dialog.component.ts
```diff
@@ -0,0 +1,69 @@
...
+      return $localize`Hours to delay is required`;
+    }
+    if ((e = errs.min)) {
+      return $localize`Specify at least ${e.min} hours`;
+    }
+    if ((e = errs.max)) {
+      return $localize`Can't exceed ${e.max} hours`;
+    }
+    if ((e = errs.pattern)) {
+      return $localize`Specify a full amount of hours for the delay`;
```
mathis-marcotte: I know what you mean here, meaning like no decimals allowed. But I don't know if "a full amount of hours" is the right way to phrase this. 

Maybe like `Specify a number of hours for the delay (no decimals)` ? (if only we could just tell users to give a non-negative integer value)

frontend/jupyter/src/app/pages/index/index-default/index-default.component.ts
```diff
@@ -260,6 +266,54 @@ export class IndexDefaultComponent implements OnInit, OnDestroy {
...
+    ref.afterClosed().subscribe(res => {
+      const config: SnackBarConfig = {
+        data: {
+          msg: ``,
+          snackType: SnackType.Success,
+        },
+        duration: 2000,
+      };
+
+      if (res === undefined || res.status === DELAY_DIALOG_RESP.CANCEL) {
```
mathis-marcotte: This can be removed if you aren't doing anything with it. Keeps the code cleaner.

frontend/jupyter/src/app/services/backend.service.ts
```diff
@@ -261,6 +261,20 @@ export class JWABackendService extends BackendService {
...
+    namespace: string,
+    name: string,
+    timehours: string,
+  ): Observable<string> {
+    const url = `api/namespaces/${namespace}/notebooks/${name}/keepalive`;
+    return this.http
+      .patch<JWABackendResponse>(url, { timehours: timehours })
+      .pipe(
+        catchError(error => this.handleError(error)),
+        map(_ => 'started'),
```
mathis-marcotte: This seems like a weird response message. Do you even need this response message? If so, maybe just use a different value like "delayed" or something? Something more accurate with the action performed instead of what seems like just the copy pasted text from the startNotebook response that you may have missed.

frontend/jupyter/src/app/services/actions.service.ts
```diff
@@ -169,6 +169,30 @@ export class ActionsService {
...
+    name: string,
+    timehours: string,
+  ): Observable<string> {
+    return new Observable(subscriber => {
+      this.backend
+        .updateKeepAlive(namespace, name, timehours)
+        .subscribe(response => {
+          const config: SnackBarConfig = {
+            data: {
+              msg: $localize`Adding auto-shutdown delay of ${timehours} hours to '${name}'...`,
```
mathis-marcotte: The word "adding" here seems a bit misplaced here. What do you think of saying `Delaying the auto-shutdown of ${name} by ${timehours} hours...` ?

frontend/jupyter/src/app/pages/index/index-default/index-default.component.ts
```diff
@@ -260,6 +266,54 @@ export class IndexDefaultComponent implements OnInit, OnDestroy {
...
+        // If we want to add any messages
+      } else {
+        if (res.status === DELAY_DIALOG_RESP.ACCEPT) {
+          config.data.msg =
```
mathis-marcotte: If you acknowledge my other comment above, then what could be done is just declare the config variable here instead of declaring it outside the if, but then only using it inside the if statement. That way, you can just declare that variable with the message you want directly. (I understand though that if you wanted that other IF condition, then it makes sense how you have it here.)

frontend/jupyter/cypress/e2e/main-page.cy.ts
```diff
@@ -274,6 +274,122 @@ describe('Main tables', () => {
...
+        .click();;
+      cy.get('.mat-mdc-dialog-title')
+        .should('be.visible')
+        .and(
+          'have.text',
+          'Delay auto-shutdown for a-test-01',
+        );
+      cy.get('[data-cy-form-button="formDelayCtrlSubmit"]').should('be.enabled');
+      cy.get('.mat-mdc-dialog-actions > button').contains('SUBMIT').click();
+      cy.intercept(
```
mathis-marcotte: You should declare the intercept before the command to click the button, if not, the call might be sent without actually being mocked (so just switch the order with the line above)

frontend/jupyter/cypress/e2e/main-page.cy.ts
```diff
@@ -274,6 +274,122 @@ describe('Main tables', () => {
...
+      cy.get('[data-cy-form-button="formDelayCtrlSubmit"]').should('be.enabled');
+      // Zero
+      cy.get('[data-cy-form-input="delayTime"]').find('input').clear();
+      cy.get('[data-cy-form-input="delayTime"]').find('input').type('0');
+      cy.get('[data-cy-form-button="formDelayCtrlSubmit"]').should('be.disabled');
+      // Max
+      cy.get('[data-cy-form-input="delayTime"]').find('input').clear();
+      cy.get('[data-cy-form-input="delayTime"]').find('input').type('72');
+      cy.get('[data-cy-form-button="formDelayCtrlSubmit"]').should('be.enabled');
+      // Negative
```
mathis-marcotte: Maybe add a decimal test as well?

frontend/jupyter/cypress/e2e/main-page.cy.ts
```diff
@@ -274,6 +274,122 @@ describe('Main tables', () => {
...
+      cy.get('[data-cy-form-button="formDelayCtrlSubmit"]').should('be.disabled');
+      // Valid
+      cy.get('[data-cy-form-input="delayTime"]').find('input').clear();
+      cy.get('[data-cy-form-input="delayTime"]').find('input').type('24');
+      cy.get('[data-cy-form-button="formDelayCtrlSubmit"]').should('be.enabled');
+      // Above max
+      cy.get('[data-cy-form-input="delayTime"]').find('input').clear();
+      cy.get('[data-cy-form-input="delayTime"]').find('input').type('73');
+      cy.get('[data-cy-form-button="formDelayCtrlSubmit"]').should('be.disabled');
+      // Valid
```
mathis-marcotte: You have two "valid" tests, you should only need 1.

frontend/jupyter/cypress/e2e/main-page.cy.ts
```diff
@@ -274,6 +274,122 @@ describe('Main tables', () => {
...
+        .find('button[data-cy-menu-icon-action="keep_alive"]')
+        .and('be.enabled')
+        .click();
+      cy.get('.mat-mdc-dialog-title')
+        .should('be.visible')
+        .and(
+          'have.text',
+          'Delay auto-shutdown for a-test-01',
+        );
+      // Min
```
mathis-marcotte: For all these tests (except for the "valid" one), it might be a good idea to also confirm that the expected error message is displayed.

For example, for the notebook form, the tests for the CPU validates something like 
```
// maximum value
      cy.get('[data-cy-form-input="cpu"]').find('input').type('40.0');
      cy.get('[data-cy-form-input="cpu"]')
        .find('input')
        .should('have.class', 'ng-invalid');
      cy.get('[data-cy-form-input="cpu"]')
        .find('mat-error')
        .should('have.text', "Can't exceed 14 CPUs");
````

File: frontend/jupyter/cypress/e2e/main-page.cy.ts [hunk 24]
```diff
@@ -274,6 +274,122 @@ describe('Main tables', () => {
         'http://localhost:4200/notebook/details/kubeflow-user/a-dog-breed-katib',
       );
     });
+
+    it('Delay menu should be disabled if notebook not ready', () => {
```
mathis-marcotte: I like the tests you added (well with the little other comments I wrote). I just had one suggestion with the titles of your tests. 

To keep with our convention, you should have the labels of your tests start with "should" and then followed by what it's testing/expecting. You already have some of that , but you have the "delay menu" in front. 

If you want to help differentiate or separate the tests for the delay dialog, you could just do a grouping called "Delay menu/delay dialog" and then have all your test cases labeled as "should blahblahblah"

So like, if you didn't know, you can do it like :
```
it('Delay menu', () => {
    it('should be disabled if notebook not ready', () => {});

    it('should cancel dialog', () => {});

    .....
});
````

Hopefully I'm being clear here. Just trying to help keep our tests organize and consistent (because as you saw, it's really big files)

**Review by mathis-marcotte — CHANGES_REQUESTED** [review]
Various things that can be addressed. Only thing I didn't look at was the messages files, as we had talked about already.

File: notebooks.go (hunk 3 above)
Comment (wg102) [reply notebooks.go]: Typo, wrong field, corrected

File: notebooks.go (hunk 5 above)
Comment (wg102) [reply notebooks.go]: Indeed

File: notebooks.go (hunk 6 above)
Comment (wg102) [reply notebooks.go]: Added

File: notebooks.go (hunk 7 above)
Comment (wg102) [reply notebooks.go:1102]: See above, the annotation was wrong.

File: notebooks.go (hunk 9 above)
Comment (wg102) [reply notebooks.go]: True enough,
I wanted to insure even if they somehow 'block' the code aka remove the disable and force the call to the function, it would catch it, but then again, a catch on the value being correctly converted into a number + simple if id indeed clearer.

File: frontend/common/kubeflow-common-lib/projects/kubeflow/src/lib/delay-dialog/delay-dialog.component.spec.ts
Comment (wg102) [reply frontend/common/kubeflow-common-lib/projects/kubeflow/src/lib/delay-dialog/delay-dialog.component.spec.ts:1]: Modified

File: frontend/common/kubeflow-common-lib/projects/kubeflow/src/lib/delay-dialog/delay-dialog.component.ts (hunk 13 above)
Comment (wg102) [reply frontend/common/kubeflow-common-lib/projects/kubeflow/src/lib/delay-dialog/delay-dialog.component.ts]: yes

File: frontend/common/kubeflow-common-lib/projects/kubeflow/src/lib/delay-dialog/delay-dialog.component.ts (hunk 14 above)
Comment (wg102) [reply frontend/common/kubeflow-common-lib/projects/kubeflow/src/lib/delay-dialog/delay-dialog.component.ts]: yes

File: frontend/common/kubeflow-common-lib/projects/kubeflow/src/lib/delay-dialog/delay-dialog.component.ts (hunk 15 above)
Comment (wg102) [reply frontend/common/kubeflow-common-lib/projects/kubeflow/src/lib/delay-dialog/delay-dialog.component.ts]: indeed the phrasing sounded wrong to me as well

File: frontend/common/kubeflow-common-lib/projects/kubeflow/src/lib/delay-dialog/delay-dialog.module.ts (hunk 12 above)
Comment (wg102) [reply frontend/common/kubeflow-common-lib/projects/kubeflow/src/lib/delay-dialog/delay-dialog.module.ts:25]: Truthfully I though about this multiple times in my coding, but it seemed simpler not to have a service.

File: frontend/jupyter/cypress/e2e/main-page.cy.ts (hunk 24 above)
Comment (wg102) [reply frontend/jupyter/cypress/e2e/main-page.cy.ts]: Yes, sections do sound good, done

File: frontend/jupyter/cypress/e2e/main-page.cy.ts (hunk 20 above)
Comment (wg102) [reply frontend/jupyter/cypress/e2e/main-page.cy.ts]: Moved it to the top of that fonction

File: frontend/jupyter/cypress/e2e/main-page.cy.ts (hunk 21 above)
Comment (wg102) [reply frontend/jupyter/cypress/e2e/main-page.cy.ts]: added

File: frontend/jupyter/cypress/e2e/main-page.cy.ts (hunk 22 above)
Comment (wg102) [reply frontend/jupyter/cypress/e2e/main-page.cy.ts]: It is to ensure that the form gets cleared of the error message

File: frontend/jupyter/src/app/pages/index/index-default/index-default.component.ts (hunk 16 above)
Comment (wg102) [reply frontend/jupyter/src/app/pages/index/index-default/index-default.component.ts]: Used negative condition to ensure the data is no those

File: frontend/jupyter/src/app/pages/index/index-default/index-default.component.ts (hunk 19 above)
Comment (wg102) [reply frontend/jupyter/src/app/pages/index/index-default/index-default.component.ts]: see above

File: frontend/jupyter/src/app/pages/notebook-page/notebook-page.component.spec.ts (hunk 11 above)
Comment (wg102) [reply frontend/jupyter/src/app/pages/notebook-page/notebook-page.component.spec.ts]: Removed

File: frontend/jupyter/src/app/services/actions.service.ts (hunk 18 above)
Comment (wg102) [reply frontend/jupyter/src/app/services/actions.service.ts]: yes sounds better

File: notebooks.go (hunk 9 above)
Comment (mathis-marcotte) [reply notebooks.go]: Yeah exactly! 
If the string -> int conversion fails, then it's most likely not a number. And then, you can just simply validate as a number, which is way easier to do than a string.

File: notebooks.go [hunk 25]
```diff
@@ -1074,39 +1073,38 @@ func (s *server) UpdateNotebookForCulling(w http.ResponseWriter, r *http.Request
...
+	notebook, err := s.listers.notebooks.Notebooks(namespaceName).Get(notebookName)
 	if err != nil {
-		fmt.Println("Error while parsing:", err)
+		s.error(w, r, err)
 		return
 	}
+
 	updatedTime := time.Now().Add(time.Duration(numKeepAliveTime) * time.Hour)
-	log.Printf("Updated time notebook %q", updatedTime);
+	log.Printf("Updated notebook %q with time notebook %q", notebookName, updatedTime);
```
mathis-marcotte: You are saying notebook twice here in this log. "time notebook" is not really a thing. 

You should make it more like "updated notebook %notebook last activity time to %time".

And also just seeing this now, this log print should probably be AFTER the update actually happens without errors, not before it, since you are saying "updated" implying that it passed.

notebooks.go
```diff
@@ -1491,26 +1489,11 @@ func validateNotebookVolume(req volrequest, validsizes map[int64]bool) error {
 	return nil
 }
 
-func validateCullingDelay(request delaycullingrequest) error {
+func validateCullingDelay(int delayHours) error {
 	var validationErrors []string
```
mathis-marcotte: This looks like a much simpler function to read after you changes. But this variable doesn't seem to be needed anymore. If that is the case, remove it.

frontend/common/kubeflow-common-lib/projects/kubeflow/src/lib/delay-dialog/delay-dialog.module.ts
mathis-marcotte: Maybe we can look at implementing a global dialog service in the feature if need be. 
For now, we can keep it as you made it, and that is something that can be refactored independently in the future if we want

frontend/jupyter/cypress/e2e/main-page.cy.ts
```diff
@@ -275,120 +275,146 @@ describe('Main tables', () => {
...
+          .and(
+            'have.text',
+            'Delay auto-shutdown for a-test-01',
+          );
+        cy.get('[data-cy-form-button="formDelayCtrlSubmit"]').should('be.enabled');
+        cy.get('.mat-mdc-dialog-actions > button').contains('SUBMIT').click();
+        cy.wait('@mockDelayCulling').its('response.statusCode').should('eq', 200);
+        cy.get('mat-dialog-container').should('not.exist');
+      });
+      it('Delay culling submit disabled if invalid data', () => {
```
mathis-marcotte: Maybe this test title can be improved to follow the convention?
With it being in the delay section already, maybe a better title would be "should disable submit if invalid data" or something?

frontend/jupyter/cypress/e2e/main-page.cy.ts
mathis-marcotte: Oh okay, that is fair then!

frontend/jupyter/cypress/e2e/main-page.cy.ts
```diff
@@ -275,120 +275,146 @@ describe('Main tables', () => {
...
+        cy.get('[data-cy-form-input="delayTime"]').find('input').type('a');
+        cy.get('[data-cy-form-button="formDelayCtrlSubmit"]').should('be.disabled');
+        cy.get('[data-cy-form-input="delayTime"]').find('input').should('have.class', 'ng-invalid');
+        cy.get('[data-cy-form-input="delayTime"]').find('mat-error')
+        .should('have.text', "Hours to delay is required");
+        // Valid - to clear the message
+        cy.get('[data-cy-form-input="delayTime"]').find('input').clear();
+        cy.get('[data-cy-form-input="delayTime"]').find('input').type('1');
+        cy.get('[data-cy-form-button="formDelayCtrlSubmit"]').should('be.enabled');
+        // Invalid - letters won't be allowed in the input
```
mathis-marcotte: Maybe a quick update on this comment to mention that this tests more for a decimal input than a letter. 
Seems like a copy pasted comment message from above in the code.

frontend/jupyter/src/app/pages/index/index-default/index-default.component.ts
```diff
@@ -283,9 +283,7 @@ export class IndexDefaultComponent implements OnInit, OnDestroy {
         duration: 2000,
       };
 
-      if (res === undefined || res.status === DELAY_DIALOG_RESP.CANCEL) {
-        // If we want to add any messages
-      } else {
+      if (!(res === undefined || res.status === DELAY_DIALOG_RESP.CANCEL)) {
```
mathis-marcotte, line 286: You are doing `res.status === CANCEL` and then inside of it, you are doing `res.status === ACCEPT`. 

That doesn't really make sense, that second if block can't ever trigger in those conditions.

mathis-marcotte (changes requested):

frontend/jupyter/src/app/services/backend.service.ts
wg102: Removed

notebooks.go
wg102: moved and updated

notebooks.go
wg102: If you're talking about the validationError, it has been removed.

frontend/jupyter/cypress/e2e/main-page.cy.ts
wg102: corrected

frontend/jupyter/cypress/e2e/main-page.cy.ts
wg102: corrected

frontend/jupyter/src/app/pages/index/index-default/index-default.component.ts
wg102: There is a "!" before the first statement. 
The second if is to catch in case there are other statuses.

mathis-marcotte: Oh right!

frontend/jupyter/i18n/fr/messages.fr.xlf
```diff
@@ -97,6 +97,46 @@
...
+        <source>Hours to delay is required</source>
+        <target>Nombre d'heures du délai requis</target>
+        <context-group purpose="location">
+          <context context-type="sourcefile">projects/kubeflow/src/lib/delay-dialog/delay-dialog.component.ts</context>
+          <context context-type="linenumber">57</context>
+        </context-group>
+      </trans-unit>
+      <trans-unit id="2488421228990292680" datatype="html">
+        <source>Specify at least <x id="PH" equiv-text="e.min"/> hour(s)</source>
+        <target>Spécifiez un minimum de <x id="PH" equiv-text="e.min"/> heures</target>
```
mathis-marcotte: Should it be "heure(s)" here? Because for 1, it would be 1 heure, unless I am wrong.

frontend/jupyter/i18n/fr/messages.fr.xlf
```diff
@@ -97,6 +97,46 @@
...
+        <source>Hours to delay</source>
+        <target>Nombre d'heures du délai</target>
+        <context-group purpose="location">
+          <context context-type="sourcefile">projects/kubeflow/src/lib/delay-dialog/delay-dialog.component.html</context>
+          <context context-type="linenumber">6,10</context>
+        </context-group>
+      </trans-unit>
+      <trans-unit id="4903387837252986467" datatype="html">
+        <source>Hours to delay is required</source>
+        <target>Nombre d'heures du délai requis</target>
```
mathis-marcotte: Ok maybe just an opinion, what about "Heures pour le délai"? And then here it would be "Heures pour le délai sont requis"? Sounds less rough to me, but what do you think?

frontend/jupyter/i18n/fr/messages.fr.xlf
```diff
@@ -97,6 +97,46 @@
...
+        <source>Can&apos;t exceed <x id="PH" equiv-text="e.max"/> hours</source>
+        <target>Ne peut pas excéder <x id="PH" equiv-text="e.max"/> heures</target>
+        <context-group purpose="location">
+          <context context-type="sourcefile">projects/kubeflow/src/lib/delay-dialog/delay-dialog.component.ts</context>
+          <context context-type="linenumber">63</context>
+        </context-group>
+      </trans-unit>
+      <trans-unit id="1652364953727905903" datatype="html">
+        <source>Specify a number of hours for the delay (no decimals)</source>
+        <target>Spécifiez le nombre d'heures du délai (chiffre entier)</target>
```
mathis-marcotte: Here is related to my other comment, what about "... nombre d'heures pour le délai... "? instead of "du délai"? But again, mainly just seeing what you think

frontend/jupyter/i18n/fr/messages.fr.xlf
```diff
@@ -1898,63 +1938,103 @@
...
         <target>Bloc-notes</target>
         <context-group purpose="location">
           <context context-type="sourcefile">src/app/pages/index/index-default/index-default.component.ts</context>
-          <context context-type="linenumber">122</context>
+          <context context-type="linenumber">125</context>
+        </context-group>
+      </trans-unit>
+      <trans-unit id="3864460974886321903" datatype="html">
+        <source>Updating last activity to add </source>
+        <target>Mise-à-jour de la dernière activité pour ajouter </target>
```
mathis-marcotte: source message and target message seem to cut off here

See comment in index-component.ts

frontend/jupyter/src/app/pages/index/index-default/index-default.component.ts
```diff
@@ -260,6 +266,52 @@ export class IndexDefaultComponent implements OnInit, OnDestroy {
...
+          msg: ``,
+          snackType: SnackType.Success,
+        },
+        duration: 2000,
+      };
+
+      if (!(res === undefined || res.status === DELAY_DIALOG_RESP.CANCEL)) {
+        if (res.status === DELAY_DIALOG_RESP.ACCEPT) {
+          config.data.msg =
+            $localize`Updating last activity to add ` + res.hours + ` hours`;
```
mathis-marcotte: Don't append with + here, it messes up the localization because it's only applying to the first part of the string. 
Use `blablah ${res.hours} blahbalh` to add variables to your string

frontend/jupyter/i18n/fr/messages.fr.xlf
```diff
@@ -1898,63 +1938,103 @@
...
+        <source>Delay auto-shutdown for <x id="PH" equiv-text="name"/></source>
+        <target>Délai de l'arrêt automatique pour <x id="PH" equiv-text="name"/></target>
+        <context-group purpose="location">
+          <context context-type="sourcefile">src/app/pages/index/index-default/index-default.component.ts</context>
+          <context context-type="linenumber">304</context>
+        </context-group>
+      </trans-unit>
+      <trans-unit id="5729881851916394567" datatype="html">
+        <source>This will keep the notebook alive for the number of hours specified</source>
+        <target>Afin de maintenir le bloc-notes actif pendant le nombre d’heures spécifié</target>
```
mathis-marcotte: Afin seems off here. What about `Ceci va maintenir le ....`

frontend/jupyter/i18n/fr/messages.fr.xlf
```diff
@@ -2214,11 +2302,11 @@
         </context-group>
         <context-group purpose="location">
           <context context-type="sourcefile">src/app/services/actions.service.ts</context>
-          <context context-type="linenumber">176</context>
+          <context context-type="linenumber">200</context>
         </context-group>
         <context-group purpose="location">
           <context context-type="sourcefile">src/app/services/config.ts</context>
-          <context context-type="linenumber">10</context>
+          <context context-type="linenumber">105</context>
```
mathis-marcotte: in the english file, this is still at "10"

frontend/jupyter/i18n/fr/messages.fr.xlf
mathis-marcotte: This line number was updated in the english file

frontend/jupyter/i18n/fr/messages.fr.xlf
```diff
@@ -2698,15 +2794,31 @@
         <target>Avertissement: Toutes données dans ce volume seront perdues.</target>
         <context-group purpose="location">
           <context context-type="sourcefile">src/app/services/actions.service.ts</context>
-          <context context-type="linenumber">175</context>
+          <context context-type="linenumber">199</context>
+        </context-group>
+      </trans-unit>
+            <trans-unit id="6943486270557756671" datatype="html">
```
mathis-marcotte, line 2800: this trans-unit doesn't seem to be added in the english file

mathis-marcotte (changes requested):

frontend/jupyter/i18n/fr/messages.fr.xlf
wg102: Le nombre d'heure(s) de délai est requis

frontend/jupyter/i18n/fr/messages.fr.xlf
wg102: Done

frontend/jupyter/i18n/fr/messages.fr.xlf
wg102: yeah I was on the fence too, it's a tad longer but it seems a bit cleaner, so will do

frontend/jupyter/i18n/fr/messages.fr.xlf
wg102: fixed

frontend/jupyter/src/app/pages/index/index-default/index-default.component.ts
wg102: fixed

frontend/jupyter/i18n/fr/messages.fr.xlf
wg102: Ceci permettra de maintenir le bloc-notes actif pendant le nombre d’heures spécifié

frontend/jupyter/i18n/fr/messages.fr.xlf
wg102: It might not be in my pr as a change but it is in the english file
image

frontend/jupyter/i18n/fr/messages.fr.xlf
wg102: Updated

frontend/jupyter/i18n/fr/messages.fr.xlf
wg102: oops, fixed. Since english was regen and is still 10

frontend/jupyter/i18n/fr/messages.fr.xlf
mathis-marcotte: Oh sorry yeah it is in the english file

frontend/jupyter/i18n/fr/messages.fr.xlf
mathis-marcotte: yep that is good

mathis-marcotte (approved):
LGTM

PR #377, opened 2026-02-16
feat: Add frontend page for editing notebooks

mathis-marcotte:
For https://jirab.statcan.ca/browse/ZONE-316

### DIFF
--- a/frontend/jupyter/i18n/fr/messages.fr.xlf
+++ b/frontend/jupyter/i18n/fr/messages.fr.xlf
@@ -787,7 +787,7 @@
       </trans-unit>
       <trans-unit id="01817bd08ce56cd2966da83498f862a5c3984cc8" datatype="html">
         <source>Requested CPUs</source>
-        <target>CPU demandés</target>
+        <target>CPU demandé(s)</target>
         <context-group purpose="location">
           <context context-type="sourcefile">src/app/pages/form/form-new/form-cpu-ram/form-cpu-ram.component.html</context>
           <context context-type="linenumber">15</context>
@@ -795,7 +795,7 @@
       </trans-unit>
       <trans-unit id="329af8132d78efd4ab83314aadb6e0ae02c71398" datatype="html">
         <source>Requested Memory Gi</source>
-        <target>Mémoire demandées, en Gi</target>
+        <target>Mémoire demandée, en Gi</target>
         <context-group purpose="location">
           <context context-type="sourcefile">src/app/pages/form/form-new/form-cpu-ram/form-cpu-ram.component.html</context>
           <context context-type="linenumber">36</context>
@@ -1199,7 +1199,7 @@
       </trans-unit>
       <trans-unit id="ae905719946782d29b51de5128df22463ad168e0" datatype="html">
         <source>Volume that will be mounted in your home directory.</source>
-        <target>Le volume qui sera monté dans votre directoire maison</target>
+        <target>Le volume qui sera monté dans votre répertoire maison</target>
         <context-group purpose="location">
           <context context-type="sourcefile">src/app/pages/form/form-new/form-workspace-volume/form-workspace-volume.component.html</context>
           <context context-type="linenumber">4</context>
--- a/frontend/jupyter/i18n/fr/messages.fr.xlf
+++ b/frontend/jupyter/i18n/fr/messages.fr.xlf
@@ -588,7 +588,7 @@
         </context-group>
         <context-group purpose="location">
           <context context-type="sourcefile">src/app/services/backend.service.ts</context>
-          <context context-type="linenumber">340</context>
+          <context context-type="linenumber">350</context>
         </context-group>
       </trans-unit>
       <trans-unit id="3862207452684082108" datatype="html">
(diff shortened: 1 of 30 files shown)


frontend/jupyter/i18n/fr/messages.fr.xlf
```diff
@@ -631,28 +631,76 @@
...
         <context-group purpose="location">
-          <context context-type="sourcefile">src/app/pages/form/form-new/form-advanced-options/form-advanced-options.component.html</context>
-          <context context-type="linenumber">12</context>
+          <context context-type="sourcefile">src/app/pages/form/form-edit/form-edit.component.html</context>
+          <context context-type="linenumber">78,80</context>
+        </context-group>
+      </trans-unit>
+      <trans-unit id="31675d6cff734605d5626261d054d8d85913ad12" datatype="html">
+        <source> CANCEL </source>
+        <target> ANNULER </target>
```
wg102, line 672: More of a comment, but some times we have it in caps, and siome times we have "Cancel" and on the UI .toUppercase. Do we want to generalise that?

frontend/jupyter/src/app/pages/form/form-edit/form-edit.component.html
```diff
@@ -0,0 +1,96 @@
...
+  >
+  </lib-title-actions-toolbar>
+
+  <!--scrollable page content-->
+  <div class="page-padding lib-flex-grow lib-overflow-auto">
+    <lib-loading-spinner *ngIf="!notebookInfoLoaded"></lib-loading-spinner>
+
+    <ng-container *ngIf="notebookInfoLoaded">
+      <form
+        novalidate
```
wg102, line 16: why?

frontend/jupyter/src/app/pages/form/form-edit/form-edit.component.ts
```diff
@@ -0,0 +1,226 @@
...
+
+    // Ensure CPU input is a string
+    if (typeof notebook.cpu === 'number') {
+      notebook.cpu = notebook.cpuLimit = notebook.cpu.toString();
+    }
+
+    // Add Gi to all sizes
+    if (notebook.memory) {
+      notebook.memory = notebook.memoryLimit =
+        notebook.memory.toString() + 'Gi';
```
wg102, line 172: Note to self Make sure we ONLY ever use Gi

frontend/jupyter/src/app/pages/form/form-edit/form-edit.component.ts
```diff
@@ -0,0 +1,226 @@
...
+  }
+
+  // Set the tooltip text based on form's validity
+  setTooltipText(form: FormGroup): string {
+    let text: string;
+    if (!form.controls.valid) {
+      text = $localize`The form contains invalid fields`;
+    }
+    return text;
+  }
```
wg102, line 191: Isn't tooltip something that shows when hovering it with mouse? That seems like a weird way to show validation fail?

frontend/jupyter/src/app/pages/form/form-edit/form-edit.component.ts
```diff
@@ -0,0 +1,226 @@
...
+
+  onSubmit() {
+    const configInfo: SnackBarConfig = {
+      data: {
+        msg: $localize`Editing Notebook...`,
+        snackType: SnackType.Info,
+      },
+    };
+    this.popup.open(configInfo);
```
wg102: I think this should be named snackbar or something to follow the rest of our code.
since the 'popups' are the dialogs not snackbars

In my brain a popup needs to be 'exited'. A snack shows up and goes away on it's own, or when you click dismiss

frontend/jupyter/src/app/shared/utils/volumes/forms.ts
```diff
@@ -10,6 +10,15 @@ import {
 import { V1PersistentVolumeClaim, V1Volume } from '@kubernetes/client-node';
 import { EXISTING_SOURCE, Volume } from 'src/app/types';
 
+// Helper values
+const mountValidators: ValidatorFn[] = [
```
wg102, line 14: Ensure that there is a tooltip that tells the 'valid paaths'

wg102 (changes requested):
Not sure if coming from other place, but it would be nice if it said what the issue it.

And for the memory, is it logical to example 2Gi if you only input 2?

frontend/jupyter/src/app/pages/form/form-edit/form-edit.component.html
```diff
@@ -0,0 +1,96 @@
...
+          [mountedVolumes]="mountedVolumes"
+        >
+        </app-form-workspace-volume>
+
+        <app-form-data-volumes
+          [readonly]="false"
+          [volsArray]="formCtrl.get('datavols')"
+          [externalName]="notebookName"
+          [mountedVolumes]="mountedVolumes"
+        ></app-form-data-volumes>
```
wg102, line 63: Coming from another place but 


It seems weird when expanding the data taht Size in gi is not populated with 16.

wg102 (changes requested):

wg102:
Images for previous comments: 
image


image

frontend/jupyter/src/app/pages/form/form-edit/form-edit.component.html
mathis-marcotte: Just copy pasted how the new notebook form was done. 
I wanted to make sure I replicated the behavior as best as I could.

frontend/jupyter/i18n/fr/messages.fr.xlf
mathis-marcotte: Maybe we would want to generalise, I am not sure what is the rule or convention with it. 
to uppercase is kinda weird if it's for static text though

frontend/jupyter/src/app/pages/form/form-edit/form-edit.component.ts
mathis-marcotte: Like the new notebook form, this is for when you hover the greyed-out launch button when an error is present in the form.
It is kind of weird to just show this message on hover, maybe we should display a red error message next to the launch/submit button when errors are in the form, but i dont know what the best thing is. We just have this current design from upstream

frontend/jupyter/src/app/pages/form/form-edit/form-edit.component.html
mathis-marcotte: We have an old ticket for fixing that.
https://jirab.statcan.ca/browse/BTIS-886

frontend/jupyter/src/app/shared/utils/volumes/forms.ts
mathis-marcotte: Not sure what you mean here. This change was just to make a const for this list of validators to reduce code duplication.

frontend/jupyter/src/app/pages/form/form-edit/form-edit.component.ts
mathis-marcotte: I think this change was copied from the new notebook form. 
But I agree with you and it should be easy to update, so I will do that and update the name of this prop for both this new edit form and the new notebook form!

frontend/jupyter/src/app/pages/form/form-edit/form-edit.component.ts
mathis-marcotte: yes, this behavior is the same between this new edit page and the new notebook page

frontend/jupyter/src/app/pages/form/form-edit/form-edit.component.ts
wg102: Ok that makes sense, i wasn't sure where it applied. Good enough since it's how it behaves elsewhere

frontend/jupyter/src/app/shared/utils/volumes/forms.ts
wg102: I meant do the user see what 'valid paths' regex is? I might be misreading the regex, but to me it means that it has to be at least partially one of those two options, was wondering if the users sees so in the UI

mathis-marcotte: I didnt change the regex, just moved it, but yes it shows
image

wg102 (approved):

PR #381, opened 2026-03-02
Add edit notebook modal

mathis-marcotte:
for https://jirab.statcan.ca/browse/ZONE-351


frontend/jupyter/src/app/pages/index/index-default/index-default.component.ts
```diff
@@ -625,7 +625,8 @@ export class IndexDefaultComponent implements OnInit, OnDestroy {
   public closePVCViewerClicked(pvc: PVCProcessedObject) {
     const closeDialogConfig: DialogConfig = {
       title: `Are you sure you want to close this viewer? ${pvc.name}`,
-      message: 'Warning: Any running processes will terminate.',
+      warning: 'Warning:',
```
wg102, line 628: Does translation exist for this?

frontend/jupyter/src/app/services/actions.service.ts
```diff
@@ -25,7 +25,7 @@ export class ActionsService {
     return new Observable(subscriber => {
       const deleteDialogConfig = getDeleteDialogConfig(name);
 
-      const ref = this.confirmDialog.open(name, deleteDialogConfig);
+      const ref = this.confirmDialog.open(deleteDialogConfig);
```
wg102, line 28: Not the right place for this comment, but should the confirmDialog be modified to just be named dialog?

wg102 (changes requested):

wg102:
should there be a "No changes were submitted, Ok, Cancel" popup?

frontend/jupyter/src/app/services/actions.service.ts
mathis-marcotte: I'm okay with it being called the confirm dialog, because this dialog's purpose is still really to confirm an action. It usually expects an action to be trigger after you confirm.

We could maybe change it to just be a generic dialog, and make the cancel/submit button optional. But I don't think we have a need right now for a basical dialog with just messages.

mathis-marcotte:
> should there be a "No changes were submitted, Ok, Cancel" popup?

I didn't feel like that was necessary. I was just handling no changes as if it was the cancel button. Because either way nothing will happen to the notebook

frontend/jupyter/src/app/pages/index/index-default/index-default.component.ts
mathis-marcotte: No. But also, we don't actually display this dialog yet, it's just code from upstream. We don't support the PVCViewer yet. 

We have a ticket to visit this feature. https://jirab.statcan.ca/browse/ZONE-265

My thought was that once we officially support this feature, then we can re-evaluate if this dialog will actually needed and we can do the translation then if necessary. 

My change here was just to adapt this `DialogConfig` with the changes I had made.

wg102 (approved):

PR #387, opened 2026-07-07
feat: Add oom msg

wg102:
https://jirab.statcan.ca/browse/ZONE-501


notebooks.go
```diff
@@ -337,9 +345,45 @@ func (s *server) getNotebookData(notebook *kubeflowv1.Notebook) (notebookrespons
 		Volumes:      volumes,
 		Labels:       notebook.Labels,
 		Metadata:     notebook.ObjectMeta,
+		IsOOMKilled:  isOOMKilled,
 	}, nil
 }
 
+func isNotebookPodOOMKilled(nb *kubeflowv1.Notebook, s *server)(bool) {
```
mathis-marcotte: for the `server` parameter, can you update this to match the other server functions? 

So like 
```
func (s *server) isNotbookPodOOMKilled(nb *kubeflowv1.Notebook)
```

And then when you use this, you do `s.isNotebookPodOOMKilled(notebook)`

notebooks.go
```diff
@@ -322,6 +323,13 @@ func (s *server) getNotebookData(notebook *kubeflowv1.Notebook) (notebookrespons
 	if req, ok := notebook.Spec.Template.Spec.Containers[0].Resources.Requests[corev1.ResourceCPU]; ok {
 		cpulimit = req.AsDec()
 	}
+
+	//get the data
+	isOOMKilled := false
+	// The following is ALWAYS true
```
mathis-marcotte: This comment seems off, what do you mean by it's "always" true? Because from what I see it can definitely return false.

notebooks.go
```diff
@@ -337,9 +345,45 @@ func (s *server) getNotebookData(notebook *kubeflowv1.Notebook) (notebookrespons
...
+		IsOOMKilled:  isOOMKilled,
 	}, nil
 }
 
+func isNotebookPodOOMKilled(nb *kubeflowv1.Notebook, s *server)(bool) {
+	notebookNameRequirement, err := labels.NewRequirement("notebook-name", selection.Equals, []string{nb.Name})
+	labelSelector := labels.NewSelector().Add(*notebookNameRequirement)
+	pods, err := s.listers.pods.Pods(nb.Namespace).List(labelSelector)
+
+	if err != nil {
```
mathis-marcotte, line 357: I think that for errors, you should look at handling them like how "createDefaultNotebook()" is doing it. 

How you currently have it, you don't seem to return the error, and this seems like a case of if there is an error that the call should fail.

notebooks.go
```diff
@@ -337,9 +345,45 @@ func (s *server) getNotebookData(notebook *kubeflowv1.Notebook) (notebookrespons
...
+		pod := pods[0]
+		resp := &podresponse{
+			APIResponseBase: APIResponseBase{
+				Success: true,
+				Status:  http.StatusOK,
+			},
+			Pod: *pod,
+		}
+		
+		if err != nil {
```
mathis-marcotte: This if block seems unnecessary considering you already check the same condition earlier in the function, without ever changing the value of "err"

frontend/jupyter/cypress/e2e/main-page.cy.ts
```diff
@@ -80,7 +80,29 @@ describe('Main tables', () => {
...
+        .find(`[data-cy-resource-table-row="Name"]`)
+        .contains('a-dog-breed-katib')
+        .parent()
+        .parent()
+        .find('app-warning-icon')
+        .find('mat-icon')
+        .should('exist');
+    });
+    
+    it('should have icon for oom', () => {
```
mathis-marcotte: I think this should mean "should not have icon for OOM", looks like it shares the same title as the previous test

frontend/jupyter/src/app/pages/index/index-default/warning-icons/warning-icon.component.ts
```diff
@@ -0,0 +1,43 @@
...
+      isWarnings = this.row.isOOMKilled;
+      if (this.row.isOOMKilled) {
+        return true;
+      }
+    }
+    return isWarnings;
+  }
+
+  public getWarningsText() {
+    //nworks with default. need to fix icon first.
```
mathis-marcotte: is this comment still relevant?

frontend/jupyter/src/app/pages/index/index-default/warning-icons/warning-icon.component.ts
```diff
@@ -0,0 +1,43 @@
+import { Component } from '@angular/core';
+import { TableColumnComponent } from 'kubeflow/lib/resource-table/component-value/component-value.component';
+
+@Component({
+  selector: 'app-warning-icon',
```
mathis-marcotte, line 5: Writing this here because I'm not sure where else.

I think you are missing a unit test file for your new component. The `.spec.ts` file. I'm thinking that should be added to match the rest of our components.

frontend/jupyter/src/app/pages/index/index-default/warning-icons/warning-icon.component.ts
```diff
@@ -0,0 +1,43 @@
...
+        return true;
+      }
+    }
+    return isWarnings;
+  }
+
+  public getWarningsText() {
+    //nworks with default. need to fix icon first.
+    var thetext = '';
+    if (this.row.hasOwnProperty('isOOMKilled')) {
```
mathis-marcotte: Do you think this block could be made a bit simpler than having a nested if statement? 

Like maybe just doing 
``` 
if (this.row.hasOwnProperty('isOOMKilled') && this.row.isOOMKilled) {
}
```

or maybe even simpler assuming it works is using the [optional chaining operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)
```
if(this.row?.isOOMKilled)
```

frontend/jupyter/src/app/pages/index/index-default/warning-icons/warning-icon.component.ts
```diff
@@ -0,0 +1,43 @@
...
+  set element(elem: any) {
+    this.row = elem;
+    if (this.showWarnings()) {
+      this.textToolTip = this.getWarningsText();
+    }
+  }
+
+  public showWarnings() {
+    var isWarnings = false;
+    if (this.row.hasOwnProperty('isOOMKilled')) {
```
mathis-marcotte: Im not sure i see the purpose of the `isWarnings` variable if all this function is doing is 1 if. This is just looking at what is there currently, but is this function designed with the idea that different warnings will be added soon? Because then it makes more sense.

But also, I think this if could be improved like my other comment mentions, with the optional chaining operator.
Like for example 
```
public showWarnings(): boolean {
    if (this.row?.isOOMKilled) {
        return true;
    }
    return false;
}
```

Or could even be done on one line with, but that may not be as good if this function is planned to be changed with new warnings added
```
public showWarnings(): boolean {
    return this.row?.isOOMKilled ? true : false;
}
```

mathis-marcotte (changes requested):
Some small things here and there that I thought could be improved

frontend/jupyter/src/app/pages/index/index-default/warning-icons/warning-icon.component.ts
wg102: It was so that the storage full message could be in the same. Let's  show if there is warnings to show or not. I am not actually sure it made sense or not..

frontend/jupyter/cypress/e2e/main-page.cy.ts
wg102: oops

frontend/jupyter/src/app/pages/index/index-default/warning-icons/warning-icon.component.ts
wg102: yes forgot that one.

frontend/jupyter/src/app/pages/index/index-default/warning-icons/warning-icon.component.ts
wg102: no

frontend/jupyter/src/app/pages/index/index-default/warning-icons/warning-icon.component.ts
wg102: fixed

notebooks.go
wg102: that was when testing... Idk how i didnèt see the comments when i checked my code.

notebooks.go
wg102: done

notebooks.go
wg102: fixed?

frontend/jupyter/src/app/pages/index/index-default/warning-icons/warning-icon.component.ts
mathis-marcotte: Yeah in that case it could see this function as simply something like this? What do you think? (I know the `isStorageFull` isn't real yet, i was just visualizing it for when you will work on that part)
```
public showWarnings(): boolean {
    if (this.row?.isOOMKilled) {
        return true;
    }
    if (this.row?.isStorageFull) {
        return true;
    }
    return false;
}
```

notebooks.go
```diff
@@ -322,6 +323,12 @@ func (s *server) getNotebookData(notebook *kubeflowv1.Notebook) (notebookrespons
 	if req, ok := notebook.Spec.Template.Spec.Containers[0].Resources.Requests[corev1.ResourceCPU]; ok {
 		cpulimit = req.AsDec()
 	}
+
+	//get the data
+	isOOMKilled := false
+	isOOMKilled = s.isNotebookPodOOMKilled(notebook);
```
mathis-marcotte: This is fine but would it be cleaner to do it in one line? Just because the default "false" value is never used here.
So like
```
//get the data
isOOMKilled := s.isNotebookPodOOMKilled(notebook);
```

notebooks.go
mathis-marcotte: I think theres some bugs here. 

First is that this function is expected to return a `bool`, but here you are returning a `(Notebook, error)`. So one thing is that the function declaration should be updated to say it returns a `(bool, error)`.

And then this return statement needs to be updated to return a bool and not a notebook, so like `return false, errors.New()`

notebooks.go
```diff
@@ -337,9 +344,40 @@ func (s *server) getNotebookData(notebook *kubeflowv1.Notebook) (notebookrespons
...
+			APIResponseBase: APIResponseBase{
+				Success: true,
+				Status:  http.StatusOK,
+			},
+			Pod: *pod,
+		}
+
+		lastState := resp.Pod.Status.ContainerStatuses[0].LastTerminationState
+
+		if (lastState.Terminated != nil){
```
mathis-marcotte: Do you think this would be cleaner in one line instead of having a nested if? 
So like 
```
if(lastState.Terminated != nil && lastState.Terminated.Reason == "OOMKilled") {
    return true;
}
```

mathis-marcotte (changes requested):

notebooks.go
wg102: fixed

frontend/jupyter/src/app/pages/index/index-default/warning-icons/warning-icon.component.ts
wg102: I actually went with your first version, ill modify when im actually coding tht enext feature

notebooks.go
wg102: done

notebooks.go
mathis-marcotte: I think you forgot this one but i can go do this change which can make a new commit to trigger the workflows that didnt run.

notebooks.go
mathis-marcotte: I think you deleted the declaration of "lastState". I can go add it back

mathis-marcotte (approved):
lgtm

PR #394, opened 2026-08-11
Add Volume Full on Volume Table

wg102:
https://jirab.statcan.ca/browse/ZONE-525


wg102:
Current visual 
image

frontend/jupyter/src/app/pages/index/columns/usage/usage.component.scss
```diff
@@ -0,0 +1,13 @@
+.size-cell {
```
mathis-marcotte: I know you copy pasted this from my older code. But with the new design with using a separate column for the icons, we don't need some of these classes anymore. 

Could you clean up this css file to just keep the class that you are using, the `.warning-icon`?

frontend/jupyter/src/app/pages/index/columns/usage/usage.component.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  constructor() {}
+
+  isUsageWarning(): boolean {
+    let roundedVal = Math.ceil(parseFloat(this.element.usage));
+    return !isNaN(roundedVal) ? roundedVal > 95 : false;
+  }
+
+  getWarningTooltipMessage(): string {
+    return $localize`Caution; this volume is more than 95% full.`;
```
mathis-marcotte: After `Caution`, should it really be a `;` or should it be a `:`? Do you know the right grammar? I personally am not sure the real answer for sure but I feel like `:` might be the right one to use.

EDIT: Actually, based on this https://www.grammarly.com/blog/punctuation-capitalization/semicolon-vs-colon-vs-dash/ , I think a `:` is correct.

frontend/jupyter/src/app/pages/index/config.ts
```diff
@@ -53,6 +54,16 @@ export const volumeTableConfig: TableConfig = {
...
     },
+    {
+      matHeaderCellDef: ``,
+      matColumnDef: 'usageicon',
+      textAlignment: 'right',
+      style: { width: '8%' },
+      value: new ComponentValue({
+        component: UsageComponent,
+      }),
+      sort: true,
```
mathis-marcotte: I think that sort should be false here, especially because i don't know if the sorting will work without giving this column a sorting function

frontend/jupyter/i18n/messages.xlf
```diff
@@ -1481,6 +1481,13 @@
           <context context-type="linenumber">26</context>
         </context-group>
       </trans-unit>
+      <trans-unit id="1801298063046788049" datatype="html">
+        <source>Caution; this volume is more than 95% full.</source>
```
mathis-marcotte: See the comment when this message comes from, but might need to be updated

frontend/jupyter/i18n/fr/messages.fr.xlf
```diff
@@ -1673,6 +1673,14 @@
           <context context-type="linenumber">26</context>
         </context-group>
       </trans-unit>
+      <trans-unit id="1801298063046788049" datatype="html">
+        <source>Caution; this volume is more than 95% full.</source>
+        <target>Attention, ce volume est rempli à plus de 95%. </target>
```
mathis-marcotte: Here it might be like the english message and might be better to use a `:` instead of a `,`. See where the english message is defined for comment about it.

mathis-marcotte (changes requested):
Just some small things to clean up here and there

frontend/jupyter/src/app/pages/index/columns/usage/usage.component.scss
wg102: done

frontend/jupyter/src/app/pages/index/columns/usage/usage.component.ts
wg102: done

frontend/jupyter/src/app/pages/index/config.ts
wg102: done

frontend/jupyter/i18n/fr/messages.fr.xlf
wg102: done

frontend/jupyter/i18n/messages.xlf
wg102: done

mathis-marcotte (approved):
lgtm

